PR #3, opened 2022-11-20
Add GitHub actions

iluuu1994:
* x86_64 uses Docker just so that the build is more easily reproducible locally
* x86_32 uses Docker to avoid dependency issues
* aarch64 uses QEMU, which unfortunately is also fairly slow
    * We do use `install`, which will create a Docker image so that dependencies don't have to be installed as part of the build

I also got a weird linker error when linking with `gcc` that doesn't happen with `clang`:

```log
 gcc -Wall -Wextra -Wno-unused-parameter -O0 -g -DIR_DEBUG=1 -DIR_TARGET_X64 -lm -o ir -lcapstone ir_main.o ir.o ir_strtab.o ir_cfg.o ir_sccp.o ir_gcm.o ir_ra.o ir_emit.o ir_load.o ir_save.o ir_emit_c.o ir_dump.o ir_disasm.o ir_gdb.o ir_perf.o ir_check.o
/usr/bin/ld: ir_disasm.o: in function `ir_disasm_branch_target':
/home/runner/work/ir/ir/ir_disasm.c:203: undefined reference to `cs_insn_group'
/usr/bin/ld: ir_disasm.o: in function `ir_disasm':
/home/runner/work/ir/ir/ir_disasm.c:344: undefined reference to `cs_open'
/usr/bin/ld: /home/runner/work/ir/ir/ir_disasm.c:350: undefined reference to `cs_option'
/usr/bin/ld: /home/runner/work/ir/ir/ir_disasm.c:354: undefined reference to `cs_option'
/usr/bin/ld: /home/runner/work/ir/ir/ir_disasm.c:414: undefined reference to `cs_malloc'
/usr/bin/ld: /home/runner/work/ir/ir/ir_disasm.c:415: undefined reference to `cs_disasm_iter'
/usr/bin/ld: /home/runner/work/ir/ir/ir_disasm.c:453: undefined reference to `cs_disasm_iter'
/usr/bin/ld: /home/runner/work/ir/ir/ir_disasm.c:575: undefined reference to `cs_free'
/usr/bin/ld: /home/runner/work/ir/ir/ir_disasm.c:655: undefined reference to `cs_close'
collect2: error: ld returned 1 exit status
```

The symbols are absolutely there (verifiable with `nm` and proven given that `clang` works fine).

I'm not sure what the `CC = aarch64-linux-gnu-gcc --sysroot=$(HOME)/php/ARM64` is needed for. Let me know and I can adjust it accordingly for `clang`.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -1,0 +1,88 @@
+name: Push
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - "**"
+permissions:
+  contents: read
+jobs:
+  x86_64:
+    name: x86_64
+    runs-on: ubuntu-22.04
+    container:
+      image: ubuntu:22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: apt
+        run: |
+          apt update -y
+          DEBIAN_FRONTEND=noninteractive apt install -y \
+            gcc \
+            libcapstone-dev \
+            libcapstone4 \
+            make \
+            php-cli
+      - name: make
+        run: make TARGET=x86_64 all
+      - name: test
+        run: make TARGET=x86_64 test-ci
+  x86:
+    name: x86
+    runs-on: ubuntu-22.04
+    container:
+      image: ubuntu:22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: apt
+        run: |
+          dpkg --add-architecture i386
+          apt update -y
+          DEBIAN_FRONTEND=noninteractive apt install -y \
+            gcc \
+            gcc-multilib \
+            libc6:i386 \
+            make \
+            php-cli \
+            wget
+      - name: Build capstone
+        run: |
+          # capstone isn't distributed for 32-bit so we need to build it manually
+          wget https://github.com/capstone-engine/capstone/archive/4.0.2.tar.gz -O capstone.tar.gz
+          tar zxvf capstone.tar.gz
+          cd capstone-4.0.2
+          ./make.sh nix32
+          ./make.sh install
+          cd ..
+          rm -rf capstone-4.0.2 capstone.tar.gz
+      - name: make
+        run: make TARGET=x86 all
+      - name: test
+        run: make TARGET=x86 test-ci
+  aarch64:
+    name: aarch64
+    runs-on: ubuntu-20.04
+    steps:
+      - uses: actions/checkout@v3
+      - uses: uraimo/run-on-arch-action@v2
+        name: QEMU
+        id: runcmd
+        with:
+          arch: aarch64
+          distro: ubuntu22.04
+          githubToken: ${{ github.token }}
+          install: |
+            apt update -y
+            DEBIAN_FRONTEND=noninteractive apt install -y \
+              gcc \
+              libc6 \
+              libcapstone-dev \
+              libcapstone4 \
+              make \
+              php-cli
+          run: |
+            make CC=gcc TARGET=aarch64 all
+            # FIXME: For some reason some of the object files are rebuilt
+            make CC=gcc TARGET=aarch64 test-ci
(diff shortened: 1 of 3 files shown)


.github/workflows/push.yml
```diff
@@ -0,0 +1,72 @@
+name: Push
```
iluuu1994, line 1: Let me know if you prefer tabs here.

Makefile
```diff
@@ -92,7 +92,7 @@ $(SRC_DIR)/ir_load.c: $(SRC_DIR)/ir.g
 $(BUILD_DIR)/ir_fold_hash.h: $(BUILD_DIR)/gen_ir_fold_hash $(SRC_DIR)/ir_fold.h $(SRC_DIR)/ir.h
 	$(BUILD_DIR)/gen_ir_fold_hash < $(SRC_DIR)/ir_fold.h > $(BUILD_DIR)/ir_fold_hash.h
 $(BUILD_DIR)/gen_ir_fold_hash: $(SRC_DIR)/gen_ir_fold_hash.c $(SRC_DIR)/ir_strtab.c
-	$(BUILD_CC) $(CFLAGS) $(LDFALGS) -o $@ $^
+	$(CC) $(CFLAGS) $(LDFALGS) -o $@ $^
```
dstogov, line 95: BUILD_CC is necessary for cross compilation. It seems you set BUILD_CC anyway, so this change is not necessary.

Makefile
```diff
@@ -26,13 +26,13 @@ ifeq (x86_64, $(TARGET))
...
-  CC= gcc -m32
-  CFLAGS += -DIR_TARGET_X86
+  CFLAGS += -m32 -DIR_TARGET_X86
   DASM_ARCH  = x86
   DASM_FLAGS =
 endif
 ifeq (aarch64, $(TARGET))
-  CC= aarch64-linux-gnu-gcc --sysroot=$(HOME)/php/ARM64
+  # FIXME: This overwrites clang from CI
+  #CC = aarch64-linux-gnu-gcc --sysroot=$(HOME)/php/ARM64
```
dstogov, line 35: I use cross compilation. It's much faster then QEMU.  This is not a blocker, just info. I may continue using cross-compilation.

.github/workflows/push.yml
dstogov: I have no idea what are the "tabs" :)

dstogov (changes requested):
Hi @iluuu1994. Thank for the help.
Please fix the BUILD_CC issue.
To fix the GCC build try to move -lcapstone -lm to the end of command line.

.github/workflows/push.yml
iluuu1994: Instead of spaces I mean 😊 You seem to be using tabs almost exclusively for indentation.

dstogov: ahh :) I don't care. spaces are good.

I use the PHP internals indentation rules.

iluuu1994:
Also, I think you might need to approve this PR to run the pipeline. There were a few segfaults in every build in my tests. See https://github.com/iluuu1994/ir/actions/runs/3506190017/jobs/5873059004.

Makefile
iluuu1994: Ah! So none of the generated code is executed at the moment, just compared? It might still make sense to keep using QEMU if that ever becomes the case, if it's not much slower. I'll try to gather some numbers.

Makefile
iluuu1994: Ah thanks, that makes sense.

dstogov:
yeah, I plan to approve and merge this. Please remove the BUILD_CC => CC chunk.

iluuu1994:
@dstogov Ah I thought you might have a button here to approve just running the pipeline as we do on php-src. But no problem, we can fix the segfaults after. I'll have time tonight to make the given changes.

iluuu1994:
>  To fix the GCC build try to move -lcapstone -lm to the end of command line.

Ugh, such a simple change, much time wasted on my side...

Should be good now, you can see here which tests are currently failing:
https://github.com/iluuu1994/ir/actions/runs/3517588981/jobs/5895531663

dstogov (approved):

dstogov:
Thanks. I'll take care about ``tests/debug/test_var-O0.irt`` test failures.
8 tests on x86 are failed expectedly (We'll need to add --XFAIL-- section).

PR #4, opened 2023-02-08
makefile: Simplify and regroup objects and targets

weltling:
A small optimization to combine common pieces in the makefile.

- Use static pattern rules for the same operation
- Group headers prerequisities into own targets
- Reduce dups like objects to be listed just once in a var, etc.

Signed-off-by: Anatol Belski <ab@php.net>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,53 +38,29 @@ ifeq (aarch64, $(TARGET))
   DASM_FLAGS =
 endif
 
-all: $(BUILD_DIR)/ir $(BUILD_DIR)/ir_test
-
-$(BUILD_DIR)/ir: $(BUILD_DIR)/ir_main.o $(BUILD_DIR)/ir.o $(BUILD_DIR)/ir_strtab.o $(BUILD_DIR)/ir_cfg.o \
+OBJS_COMMON = $(BUILD_DIR)/ir.o $(BUILD_DIR)/ir_strtab.o $(BUILD_DIR)/ir_cfg.o \
 	$(BUILD_DIR)/ir_sccp.o $(BUILD_DIR)/ir_gcm.o $(BUILD_DIR)/ir_ra.o $(BUILD_DIR)/ir_emit.o \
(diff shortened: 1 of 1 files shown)


dstogov (changes requested):
The patch breaks ``make TARGET=x86_64 BUILD_DIR=build_x86_64``

weltling:
Thanks for the check. Looks like it is broken already in today's master 9b6b6996c48d1abd00c3bfb977698951e2460148 already, running this:

```
$ make TARGET=x86_64 BUILD_DIR=build_x86_64
gcc -Wall -Wextra -Wno-unused-parameter -O0 -g -DIR_DEBUG=1 -DIR_TARGET_X64 -o build_x86_64/ir_main.o -c ir_main.c
Assembler messages:
Fatal error: can't create build_x86_64/ir_main.o: No such file or directory
make: *** [Makefile:85: build_x86_64/ir_main.o] Error 1
```

But sure I can make it work in the scope of this PR.

Thanks

dstogov:
> Fatal error: can't create build_x86_64/ir_main.o: No such file or directory

Do you have ``build_x86_64`` directory?

weltling:
Ah, i see, so it was expected to be pre-created manually. I fixed this in the latest commit, also it'll create the target dir from the makefile and remove on `clean`.

Thanks

dstogov (changes requested):
Something is still wrong. For some reason ``BUILD_DIR=build_x86_64 TARGET=x86_64`` recompiles all object files even if they are already exists.

weltling:
Thanks for the further check. That's the BUILD_DIR auto creation, it was wrongly put as a prerequisite for objects which changed the dir timestamp every time. Moved that to be only a dep for the `all` target, that seems to fix it. I've also rebased to the latest and squashed all the history following the discussion. One could also refrain from the build dir creation for now, should it cause issues..

Thanks

dstogov (approved):

PR #5, opened 2023-02-12
Fix incorrect val.i64 check

nielsdos:
Fixes a couple of typos in the check condition:
* `0x7ffffff` -> `0x7fffffff` (one f too little)
* `&&` means that the check is always false, should've been `||`
* `>=` -> `>` because 0x7ffffff is a valid i32.


dstogov (approved):
The fix is definitely right. Thanks.
In general, we may also accept negative ``int32_t`` numbers.

PR #6, opened 2023-02-12
Fix SCCP use list update for op3

nielsdos:
There's a typo: op2 was checked twice but op1 was *not* checked.

Found using static analysis.


dstogov (approved):
Many thanks to your static analyser. What do you use?

nielsdos:
I used PVS studio for this one. It is available for free to use for open source software.

PR #7, opened 2023-02-18
Fix temporary register allocation for IR_STORE_INT

arnaud-lb:
Found while exploring / testing.

ir_get_target_constraints() mistakenly tests the instruction type and value instead of the operands'.

This results in an assertion failure in ir_emit_store_int() when op2 is a non-32bit const because no register was allocated for op2.


dstogov (approved):
Good catch. Thanks.

PR #8, opened 2023-02-22
build: MSVC compatibility

weltling:
This patch brings inital MSVC support, adding very basic fixes for code and build infra. Changes included:

- Basic code fixes to allow compiling with MSHV
- Makefile for nmake
- Extended pipeline yaml for building on Windows
- The first simple test using `ir.exe test.ir` is executed, not all the code parts are tested yet

The dependency on `libcapstone` is provided through [vcpkg](https://vcpkg.io/).

There is more to go especially to implement the full test suite and also involve the runtime execution part. There will be more build, code, test and infrastructure work required which can be build upon the base brought in this patch.

Perhaps this patch should be seen as a discussion starter. A cleaner approach could be also to separate the MSVC related code into some framework under `./win32/...` instead of polluting the codebase with `#ifdef`s (and more are to come). That way, there could be say also OS code separated like Linux/Unix/Windows specific that would share some headers and provide a platform specific code in separate C files. Maybe there can be a better approach, it just looks to me that an initial MSVC support is a good point to work out some long term strategy for a cleaner code base.

Signed-off-by: Anatol Belski <ab@php.net>

### DIFF
--- a/ir.c
+++ b/ir.c
@@ -18,7 +18,12 @@
 # define _GNU_SOURCE
 #endif
 
-#include <sys/mman.h>
+#ifndef _WIN32
+# include <sys/mman.h>
+#else
+# define WIN32_LEAN_AND_MEAN
+# include <windows.h>
+#endif
 
 #include "ir.h"
 #include "ir_private.h"
@@ -1125,7 +1130,7 @@ void ir_hashtab_init(ir_hashtab *tab, uint32_t size)
 void ir_hashtab_free(ir_hashtab *tab)
 {
 	uint32_t hash_size = (uint32_t)(-(int32_t)tab->mask);
-	char *data = tab->data - (hash_size * sizeof(uint32_t));
+	char *data = (char*)tab->data - (hash_size * sizeof(uint32_t));
 	ir_mem_free(data);
 	tab->data = NULL;
 }
@@ -1209,6 +1214,49 @@ void ir_hashtab_key_sort(ir_hashtab *tab)
 }
 
 /* Memory API */
+#ifdef _WIN32
+void *ir_mem_mmap(size_t size)
+{
+	void *ret;
+
+#ifdef _M_X64
+	DWORD size_hi = size >> 32, size_lo = size & 0xffffffff;
+#else
+	DWORD size_hi = size, size_lo = 0;
+#endif
+
+	HANDLE h = CreateFileMapping(INVALID_HANDLE_VALUE, NULL, PAGE_EXECUTE_READWRITE, size_hi, size_lo, NULL);
+
+	ret = MapViewOfFile(h, FILE_MAP_COPY | FILE_MAP_EXECUTE, size_hi, size_lo, size);
+	if (!ret) {
+		CloseHandle(h);
+	}
+
+	return ret;
+}
+
+int ir_mem_unmap(void *ptr, size_t size)
+{
+	/* XXX file handle is leaked. */
+	UnmapViewOfFile(ptr);
+	return 1;
+}
+
+int ir_mem_protect(void *ptr, size_t size)
+{
+	return 1;
+}
+
+int ir_mem_unprotect(void *ptr, size_t size)
+{
+	return 1;
+}
+
+int ir_mem_flush(void *ptr, size_t size)
+{
+	return 1;
+}
+#else
 void *ir_mem_mmap(size_t size)
 {
 	return mmap(NULL, size, PROT_EXEC, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
@@ -1242,6 +1290,7 @@ int ir_mem_flush(void *ptr, size_t size)
 #endif
 	return 1;
 }
+#endif
 
 /* Alias Analyses */
 typedef enum _ir_alias {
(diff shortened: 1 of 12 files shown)


ir.c
```diff
@@ -325,7 +329,7 @@ void ir_init(ir_ctx *ctx, ir_ref consts_limit, ir_ref insns_limit)
 
 	ctx->strtab.data = NULL;
 
-	buf = ir_mem_malloc((consts_limit + insns_limit) * sizeof(ir_insn));
+	buf = ir_mem_calloc(consts_limit + insns_limit, sizeof(ir_insn));
```
dstogov, line 332: calloc() may be more expensive as it needs to clear the memory

ir.h
```diff
@@ -338,9 +357,7 @@ typedef union _ir_val {
 			uint32_t                   u32;
 			int32_t                    i32;
 			float                      f;
-#ifndef IR_64
-            uintptr_t                  addr;
-#endif
+			ADDR_MEMBER
```
dstogov, line 360: Why do we need this additional ADDR_MEMBER macro?

ir.h
```diff
@@ -374,7 +392,7 @@ typedef struct _ir_insn {
 				union {
 					IR_STRUCT_LOHI(
 						uint8_t        op,
-						uint8_t        type;
+						uint8_t        type
```
dstogov, line 395: This ``;`` should probably be added in the ``IR_STRUCT_LOHI`` macro

ir_private.h
```diff
@@ -133,9 +143,9 @@ IR_ALWAYS_INLINE uint32_t ir_ntzl(uint64_t num)
 	unsigned long index;
 
 #if defined(_WIN64)
-	if (!BitScanForward64(&index, num)) {
+	if (!_BitScanForward64(&index, num)) {
 #else
-	if (!BitScanForward(&index, num)) {
+	if (!_BitScanForward(&index, num)) {
 #endif
```
dstogov, line 149: I'm not sure if ``BitScanForward()`` works for ``uint64_t``. I know, this code was here before this patch.

ir_x86.h
```diff
@@ -87,7 +87,11 @@ enum _ir_reg {
 #define IR_REG_GP_LAST  (IR_REG_FP_FIRST - 1)
 #define IR_REG_FP_LAST  (IR_REG_NUM - 1)
 
-#define IR_REGSET_64BIT 0
+#ifdef _M_X64 /* MSVC */
+# define IR_REGSET_64BIT 1
+#else
+# define IR_REGSET_64BIT 0
+#endif
```
dstogov, line 94: I think we don't need 64-bits for regset (32-bits should be enough). The number of CPU registers on Linux x86_64 and Windows 64 are the same.

dstogov (changes requested):
Looks fine. See the minor notes.

ir_x86.h
weltling: The reason to change this was other warnings like below, which signify issues with a regression potential:

`.\ir_ra.c(488): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)`

I'll dig more on this, check to revert and and perhaps try to fix these cases locally at where they're reported.

Thanks

ir.c
weltling: Yeah, `calloc` is definitely more expensive. It was in an early version of the patch where it seemed some members of `ctx->ir_base` were not initialized to zero and that might have caused some failures. I've reverted this and seems it doesn't reproduce the issues i've initially seen. So keep it reverted. 

Thanks

ir.h
weltling: It's MSVC that refuses to compile this piece. In the current version of the code the `#ifdef` is used inside of a `IR_STRUCT_LOHI` macro call, so seems refusing to compile it is sane. What happens there I think is that it would take the preprocessor directives and put them as C without actually interpreting the `#ifdef`, so that would cause a syntax error.

Thanks

ir.h
weltling: This seems to have been a typo in the original variant. As far as i can see, it's always used with a semicolon at the end, like `IR_STRUCT_LOHI(...);`. It's that the second semicolon after the second member that seems to have made issue with MSVC. 

Thanks

ir_private.h
weltling: Perhaps it's not that bad if the values are expected to be in the range. But depends on what is passed in here, maybe it's indeed better to have a different signature. I'll verify more on this.

Thanks

weltling: I made some ifdef at another place also, for `ir_nlz`, as 64-bit version doesn't exist in x86 build. Perhaps it'd be cleaner, for 64-bit input, to just have the plain C implementation. Maybe it's a better way to do that.

Thanks

ir.h
dstogov: right. I'm not sure how this works with GCC :)

ir_private.h
dstogov: I think, use clean C version for uint64_t on 32-bit Windows is  acceptable. It shouldn't be used at all.

ir.h
dstogov: Yeah. This is my problem. I can take care after this is merged.

Issue #12, opened 2023-03-13
amd64: Use CPUID  to automatically enable features


weltling:
Certain features can be enabled automatically, based on the data advertised through CPUID. There needs to be a framework to

- Retrieve the processor capabilities
- Provide API to consuming parts within `IR` to self modify the generated code depending on processor features

The feature enablement can be done an the library level or provided to the consumer for further decision.

PR #21, opened 2023-04-10
[RFC] cpu: Add framework for CPU feature handling

weltling:
This comes with the initial feature to query CPUID on x86. Supported are both GCC (like) based build and MSVC builds targeting Windows. 

For now, all the implementation would fit just in the header. Only the necessary leafs are saved for the further querying. There is an initial set of macros like `ir_cpu_have_avx()` provided to query features. These are just prepared, more can be added right away to support a wider choice out of the box.

There might be reasons to move some part of the code into a C file. That might simplify the API by hiding the masks struct in a C file. However, hiding things in a C file might require to have some more overhead with regard to function calls.

Fixes: #12

Signed-off-by: Anatol Belski <anbelski@linux.microsoft.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,8 @@ endif
 OBJS_COMMON = $(BUILD_DIR)/ir.o $(BUILD_DIR)/ir_strtab.o $(BUILD_DIR)/ir_cfg.o \
 	$(BUILD_DIR)/ir_sccp.o $(BUILD_DIR)/ir_gcm.o $(BUILD_DIR)/ir_ra.o $(BUILD_DIR)/ir_emit.o \
 	$(BUILD_DIR)/ir_load.o $(BUILD_DIR)/ir_save.o $(BUILD_DIR)/ir_emit_c.o $(BUILD_DIR)/ir_dump.o \
-	$(BUILD_DIR)/ir_disasm.o $(BUILD_DIR)/ir_gdb.o $(BUILD_DIR)/ir_perf.o $(BUILD_DIR)/ir_check.o
+	$(BUILD_DIR)/ir_disasm.o $(BUILD_DIR)/ir_gdb.o $(BUILD_DIR)/ir_perf.o $(BUILD_DIR)/ir_check.o \
+	$(BUILD_DIR)/ir_cpuinfo.o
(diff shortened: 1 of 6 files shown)


dstogov (changes requested):
Hi. This looks good, but I would like you to make a few changes.

``ir_cpu.h`` should be renamed into ``ir_cpuinfo.c``.
It should provide a single public function ``uint32_t ir_cpuinfo(void)``.
It should return a bit-set where each bit defines some CPU feature. The bits definition should be done at the end of  ``ir_x86.h``. Like the following...

```c
#define IR_X86_SSE2  (1<<0)
#define IR_X86_SSE4  (1<<1)
#define IR_X86_AVX   (1<<2)
#define IR_X86_AVX2  (1<<3)
...
```

``uint32_t ir_cpu_info(void)`` declaration should be added into ``ir.h``.

``ir_test.c`` should be made compatible with the changes. This is not very important, because once you provide these changes, I'll replace ``IR_AVX`` usage by ``IR_X86_AVX`` in a new ``ir_ctx.mflags``. I'll also take care about ``ir_main.c`` chqnges.

weltling:
I followed your comments and reorganized the code correspondingly. I also reworked the flag set preliminary to contain all `SSE` and `AVX(2)`. There will be sure more interesting flags, I just didn't want to add things without knowing whether they're immediately needed. Later the `AVX512*` and some other features as in server class CPUs might be of interest.

I found this one in the Intel docs:

https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/may-i-use-cpu-feature.html

which shows possible flags that could be used. The full set might nee a 64 bit variable.

Probably not all of the flags mentioned there in the doc will be needed, but most likely 32 bit might be good enough for what matters for `IR` itself. We might also still provide the defs for `ir_cpuid` in say `ir_x86.h`, so then it can be used for anything else, too. For example, a user could want to use features like `SHA` or `AES` or any other arbitrary feature in the consuming code, but these are most likely of no interest to be present in the mask returned by `ir_cpuinfo()`.
 
Thanks

dstogov (approved):
Now this looks fine.

Issue #14, opened 2023-03-13
frontend: Implement C++ bindings


weltling:
Based on `ir.h`, `ir_test.h` and `ir_test.c` example implementation, develop a C++ wrapper for the IR functionality. Yet it is to be started in a simple manner to make the C++ code as simple as possible, while keeping in mind to prospectively modernize for the C++11 and onwards.

PR #22, opened 2023-04-10
API: Implement C++ wrapper

weltling:
The idea behind the implementation is to get the full extent of the powerful C++ features like the strict type handling, method overloading, etc. At the same time, the C++ wrapper should not add any significant overhead to the IR code generation time, while being embeddable well into modern C++ code.

The most of the IR functions use `ir_ref` which is typeless on the C/C++ side of things. Many functions use the type constants like `IR_BOOL` and suffixes in the function names to signify the data types. The C++ paradigms allow to simplify this by wrapping types and values in a class, which can be used to overload a method handling a particular operand type.

For example, an operation like getting an absolute value using the C functionality could lead to a function API similar to: 

```
ir_ref abs_i8(ir_ctx *ctx, ir_ref op1) { return ir_fold1(ctx, IR_OPT(IR_ABS, IR_I8), op1);
ir_ref abs_i8(ir_ctx *ctx, ir_ref op1) { return ir_fold1(ctx, IR_OPT(IR_ABS, IR_I16), op1);
.....
```

where in C++, the same could be done by an approach similar to

```
i8 abs(ir_ctx *ctx, i8 op1) { return ir_fold1(ctx, IR_OPT(IR_ABS, IR_I8), op1);
i16 abs(ir_ctx *ctx, i16 op1) { return ir_fold1(ctx, IR_OPT(IR_ABS, IR_I16), op1);
.....
```
As illustrated here, in C++ the function signatures can be simplified, using the type checking and function overloading paradigms. Further simplifications, for example encapsulating the context argument by using OOP, automatically recognizing whether there's a value, ref or type, can be done.

This PR approaches an initial C++ wrapper  implementation based on the C++ paradigms as above and more. There are still some open points with regard to the data types in certain functions, possible overloading with arguments of mixed types that will need a closer attention. This PR is in first place to open the discussion on the wrapper approach, to see how for it can lead and catch any possible issues in the boiler plate code.

Fixes: #14 

Signed-off-by: Anatol Belski [anbelski@linux.microsoft.com](mailto:anbelski@linux.microsoft.com)


dstogov (commented):
The ``constval()`` is really over-designed.
The following patch simplifies implementation and usage (in the test)

```diff
diff --git a/c++/include/ir.hxx b/c++/include/ir.hxx
index 0e209cf..c1b752f 100644
--- a/c++/include/ir.hxx
+++ b/c++/include/ir.hxx
@@ -61,19 +61,16 @@ namespace ir
 
 	// i8, i16, i32, i64, u8, u16, u32, u64, addr, c, b, d, f
 #define IR_TYPE_CLS(name, cxx_type, field, flags) \
-	class field : public type, public ref { \
-	protected: \
-		cxx_type v; \
+	class field : public ref { \
 	public: \
 		static const field t; \
-		field() : v(0), type(IR_ ## name), ref(0) {} \
-		field(cxx_type _v) : v(_v), type(IR_ ## name), ref(0) {} \
-		field(const ref& _r) : v(0), type(IR_ ## name), ref(_r) {} \
-		operator cxx_type() const { return v; } \
-		operator ref() const { return r; } \
-		operator type() const { return t; } \
+		field() : ref(0) {} \
+		field(ir_ref _r) : ref(_r) {} \
+		field(const ref& _r) : ref(_r) {} \
+		operator ir_type() const { return IR_ ## name; } \
 	}; \
-	const field field::t = field{ IR_ ## name };
+	const field field::t = field{ IR_ ## name }; \
+	const auto name = IR_ ## name;
 	IR_TYPES(IR_TYPE_CLS)
 #undef IR_TYPE_CLS
 	class none : public type {
@@ -128,24 +125,28 @@ namespace ir
 		builder(uint32_t flags, int32_t consts_limit = IR_CONSTS_LIMIT_MIN, int32_t insns_limit = IR_INSNS_LIMIT_MIN)
 			: engine(flags, consts_limit, insns_limit) {}
 
-		i8 constval(i8 v) { return ref(ir_const_i8(&ctx, v)); }
-		i16 constval(i16 v) { return ref(ir_const_i16(&ctx, v)); }
-		i32 constval(i32 v) { return ref(ir_const_i32(&ctx, v)); }
-		i64 constval(i64 v) { return ref(ir_const_i64(&ctx, v)); }
-		u8 constval(u8 v) { return ref(ir_const_u8(&ctx, v)); }
-		u16 constval(u16 v) { return ref(ir_const_u16(&ctx, v)); }
-		u32 constval(u32 v) { return ref(ir_const_u32(&ctx, v)); }
-		u64 constval(u64 v) { return ref(ir_const_u64(&ctx, v)); }
-		b constval(b v) { return ref(ir_const_bool(&ctx, v)); }
-		c constval(c v) { return ref(ir_const_char(&ctx, v)); }
-		f constval(f v) { return ref(ir_const_float(&ctx, v)); }
+		i8 const_i8(int8_t v) { return ref(ir_const_i8(&ctx, v)); }
+		i16 const_i16(int16_t v) { return ref(ir_const_i16(&ctx, v)); }
+		i32 const_i32(int32_t v) { return ref(ir_const_i32(&ctx, v)); }
+		i64 const_i64(int64_t v) { return ref(ir_const_i64(&ctx, v)); }
+		u8 const_u8(uint8_t v) { return ref(ir_const_u8(&ctx, v)); }
+		u16 const_u16(uint16_t v) { return ref(ir_const_u16(&ctx, v)); }
+		u32 const_u32(uint32_t v) { return ref(ir_const_u32(&ctx, v)); }
+		u64 const_u64(uint64_t v) { return ref(ir_const_u64(&ctx, v)); }
+		b const_bool(bool v) { return ref(ir_const_bool(&ctx, v)); }
+		c const_char(char v) { return ref(ir_const_char(&ctx, v)); }
+		f const_float(float v) { return ref(ir_const_float(&ctx, v)); }
+		d const_double(double v) { return ref(ir_const_double(&ctx, v)); }
+		addr const_addr(uintptr_t v) { return ref(ir_const_addr(&ctx, v)); }
+		addr const_func_addr(uintptr_t v, uint16_t flags) { return ref(ir_const_func_addr(&ctx, v, flags)); }
+
+		addr unique_const_addr(uintptr_t v)  { return ref(ir_unique_const_addr(&ctx, v)); }
+
+		b constval(bool v) { return ref(ir_const_bool(&ctx, v)); }
+		c constval(char v) { return ref(ir_const_char(&ctx, v)); }
 		f constval(float v) { return ref(ir_const_float(&ctx, v)); }
-		d constval(d v) { return ref(ir_const_double(&ctx, v)); }
 		d constval(double v) { return ref(ir_const_double(&ctx, v)); }
-		addr constval(addr v) { return ref(ir_const_addr(&ctx, v)); }
-		ref constval(ref v) { return ir_const_str(&ctx, v); } // XXX have type str?
-		addr constval(addr v, uint16_t flags) { return ref(ir_const_func_addr(&ctx, v, flags)); }
-		ref constval(ref v, uint16_t flags) { return ir_const_func(&ctx, v, flags); } // XXX have type func?
+		addr constval(const void *v) { return ref(ir_const_addr(&ctx, (uintptr_t)v)); }
 
 		b eq(ref op1, ref op2) { return ref(ir_fold2(&ctx, IR_OPT(IR_EQ, IR_BOOL), ref(op1), ref(op2))); }
 		b ne(ref op1, ref op2) { return ref(ir_fold2(&ctx, IR_OPT(IR_NE, IR_BOOL), ref(op1), ref(op2))); }
@@ -511,7 +512,7 @@ namespace ir
 			// NOTE There is likely no chance to verify whether the number is correct.
 			//      Numbers can be passed in an arbitrary order.
 			num = num > 0 ? num : ++param_pos_cnt;
-			return ref(_ir_PARAM(&ctx, type(typ), name.c_str(), num));
+			return ref(_ir_PARAM(&ctx, ir_type(typ), name.c_str(), num));
 		}
 
 		template <class T>
diff --git a/c++/ir_test3.cxx b/c++/ir_test3.cxx
index 6a760c8..ce356d3 100644
--- a/c++/ir_test3.cxx
+++ b/c++/ir_test3.cxx
@@ -58,14 +58,14 @@ int main(int argc, char** argv) {
 	auto ci = bld.copy(x);
 	auto zi = bld.copy(bld.constval(.0));
 	auto zr = bld.copy(bld.constval(.0));
-	auto i = bld.copy(bld.constval(ir::i32{0}));
+	auto i = bld.copy(bld.const_i32(0));
 
 	auto loop = bld.loop_begin();
 		ir::d zi_1 = bld.phi(zi);
 		ir::d zr_1 = bld.phi(zr);
 		ir::i32 i_1 = bld.phi(i);
 
-		auto i_2 = bld.add(i_1, bld.constval(ir::i32{1}));
+		auto i_2 = bld.add(i_1, bld.const_i32(1));
 		auto temp = bld.mul(zr_1, zi_1);
 		auto zr2 = bld.mul(zr_1, zr_1);
 		auto zi2 = bld.mul(zi_1, zi_1);
@@ -75,9 +75,9 @@ int main(int argc, char** argv) {
 			bld.if_true(if_1);
 				bld.ret(i_2);
 			bld.if_false(if_1);
-				auto if_2 = bld.if_cond(bld.gt(i_2, bld.constval(ir::i32{1000})));
+				auto if_2 = bld.if_cond(bld.gt(i_2, bld.const_i32(1000)));
 				bld.if_true(if_2);
-					bld.ret(bld.constval(ir::i32{0}));
+					bld.ret(bld.const_i32(0));
 				bld.if_false(if_2);
 					auto loop_end = bld.loop_end();
 
```

dstogov:
Type usage in ``parm`` and ``call`` also looks more complex then in C.

The rest builder API (e.g. ``add``, ``sub``, etc) benefits from C++ type overloading, but note that this overloading works only if we know the type(s) statically. This is a serious limitation. For example in PHP JIT we keep an array of ``ir_ref``s to nodes that represent the current values of PHP variables (some of them DOUBLE others are LONG). Static typing won't work in this case. See https://github.com/dstogov/php-src/blob/php-ir/ext/opcache/jit/zend_jit_ir.c#L1147 . Using object with virtual method is also not an option.

Finally, it would be great to have C and C++ APIs without big differences, to have a common reference manual.

c++/include/ir.hxx
```diff
@@ -0,0 +1,603 @@
...
+ * (C++ wrapper for IR library)
+ * Copyright (C) 2023 by IR project.
+ * Authors: Anatol Belski <anbelski@linux.microsoft.com>
+ */
+
+#ifndef IR_HXX
+#define IR_HXX
+
+#include "ir.h"
+#include "ir_builder.h"
```
dstogov, line 12: Can we include these files under ``ir::c`` namespace to hide C declarations from C++ users.

```c++
namespace ir {
  namespace c {
#  include "ir.h"
#  include "ir_builder.h"
  }
    const auto consistency_check = ir::c::ir_consistency_check;
```

weltling:
After further consideration it became obvious, that the approach as in this PR isn't suitable and doesn't reflect concepts involved in the low level implementation. Some part of the code in this PR could be reused in the further work, but in general the approach needs to be chosen more deliberately. Thus, herewith is this PR abandoned.

Thanks a lot for your guidance and reviews, @dstogov!

PR #24, opened 2023-04-17
Enhance ir test framework

stkeke:
This PR is a series enhancement to ir-test to facilitate debugging and development.
7a62d05 [ir-test]: Use g++-12 to build ir-test
8878ddb [ir-test]: Enable user to specify tests to run
201dee1 [ir-test]: add --help option and print program usage
3a2ab53 [ir-test]: Make ir-test run from any directory

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,10 +91,10 @@ test: $(BUILD_DIR)/ir $(BUILD_DIR)/ir-test
 	$(BUILD_DIR)/ir $(SRC_DIR)/test.ir --dump --save 2>$(BUILD_DIR)/test.log
 	$(BUILD_DIR)/ir $(SRC_DIR)/test.ir --dot $(BUILD_DIR)/ir.dot
 	dot -Tpdf $(BUILD_DIR)/ir.dot -o $(BUILD_DIR)/ir.pdf
-	BUILD_DIR=$(BUILD_DIR) SRC_DIR=$(SRC_DIR) $(BUILD_DIR)/ir-test
+	$(BUILD_DIR)/ir-test $(SRC_DIR)/tests
 
 test-ci: $(BUILD_DIR)/ir $(BUILD_DIR)/ir-test
-	BUILD_DIR=$(BUILD_DIR) SRC_DIR=$(SRC_DIR) $(BUILD_DIR)/ir-test --show-diff
+	$(BUILD_DIR)/ir-test --show-diff $(SRC_DIR)/tests
 
 clean:
 	rm -rf $(BUILD_DIR)/ir $(BUILD_DIR)/ir_test $(BUILD_DIR)/*.o \
--- a/ir-test.cxx
+++ b/ir-test.cxx
@@ -12,8 +12,10 @@
 #include <sstream>
 #include <filesystem>
 #include <array>
+#include <set>
 #include <exception>
 #include <cstdlib>
+#include <cassert>
 
 #ifdef _WIN32
 # include <windows.h>
@@ -37,7 +39,8 @@ std::vector<std::string> bad_list;
 std::vector<std::tuple<std::string, std::string>> failed;
 std::vector<std::tuple<std::string, std::string, std::string>> xfailed_list;
 bool show_diff = false, colorize = true;
-std::string src_dir, build_dir, test_dir, ir_exe, ir_target;
+std::string ir_exe, ir_target;
+std::string this_exe, this_exe_path;
 
 namespace ir {
 	decltype(auto) trim(std::string s) {
@@ -58,12 +61,6 @@ namespace ir {
 			result += buffer.data();
 		}
 		return result;
-	}
-
-	decltype(auto) get_dir_from_env(const char* name) {
-		const char* _tmp = std::getenv(name);
-		std::string ret = _tmp ? _tmp : ".";
-		return ret;
 	}
 
 	enum color { GREEN, YELLOW, RED };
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -85,7 +86,7 @@ $(OBJS_COMMON) $(OBJS_IR) $(OBJS_IR_TEST): $(BUILD_DIR)/$(notdir %.o): $(SRC_DIR
 	$(CC) $(CFLAGS) -I$(BUILD_DIR) -o $@ -c $<
 
 $(BUILD_DIR)/ir-test: $(SRC_DIR)/ir-test.cxx
-	$(CXX) -O3 -std=c++17 $(SRC_DIR)/ir-test.cxx -o $(BUILD_DIR)/ir-test
+	$(CXX12) -O3 -std=c++17 $(SRC_DIR)/ir-test.cxx -o $(BUILD_DIR)/ir-test
```
dstogov, line 89: I'm not sure. ``g++`` works fine for me, and I don't have ``g++-12`` installed.

ir-test.cxx
```diff
@@ -231,25 +233,79 @@ namespace ir {
...
 		} else if (!std::string(argv[i]).compare("--no-color")) {
 			::colorize = false;
+			continue;
+		} else if (!std::string(argv[i]).compare("--help")) {
+			print_help();
+			return 0;
+		} else {
+			// user specified test folders/files
+			std::string file = argv[i];
+			if (std::filesystem::exists(file)
```
dstogov, line 267: I would recommend to check for ``--`` prefix and report about unsupported option before trying argument as a file path.

ir-test.cxx
```diff
@@ -62,7 +64,7 @@ namespace ir {
 
 	decltype(auto) get_dir_from_env(const char* name) {
 		const char* _tmp = std::getenv(name);
-		std::string ret = _tmp ? _tmp : ".";
+		std::string ret = _tmp ? _tmp : ::this_exe_path;
```
dstogov, line 67: I think, since we have an ability to specify tests through arguments, we may completely remove the old way (passing arguments through environment variables). Then ``::this_exe_path`` is going to be useless too.

dstogov (changes requested):
I would accept second and third commits.

@weltling do you have any thoughts?

Makefile
weltling: Passing like `make CXX=g++-12` seems fine enough. On some distro a symlink to `g++` might not be created automatically, or there could be multiple versions. Having the compiler version hardcoded like this seems not future proof.

Thanks

ir-test.cxx
```diff
@@ -231,25 +233,79 @@ namespace ir {
...
+
+	// get test files, either specified by user or all tests by default
 	std::vector<std::string> irt_files;
+	if (user_files.empty()) {
+		// get all tests
+		::test_dir = ::src_dir + PATH_SEP + "tests";
+		ir::find_tests_in_dir(::test_dir, irt_files);
+	} else {
+		// get user specified tests
+		for (auto f : user_files) {
```
weltling, line 298: Could this loop be packed into `find_tests_in_dir`? Then it would accept a `set` as the first argument instead of a `string`. The `set` would consist either on the user supplied list of filenames, or on the `test_dir` if none given. Or maybe just use another overloaded version of `find_tests_in_dir`, that would call itself with the directory/file path, depending on what the current item in `set` is? This would lay a base for a further improvement in the future to add some kind of `glob`, too.

Thanks

ir-test.cxx
stkeke: will add this check.

ir-test.cxx
stkeke: will remove and refactor this.

ir-test.cxx
stkeke: good idea. I will refactor this code part.

Makefile
stkeke: Reverted commit 'Use g++-12 to build ir-test'

ir-test.cxx
stkeke: removed all the code part getting path from environment variable

ir-test.cxx
stkeke: packed the loop into `find_tests_in_dir` and now this function accepts a `set` as its first argument.

ir-test.cxx
stkeke: image

ir-test.cxx
```diff
@@ -218,38 +215,102 @@ namespace ir {
...
+
+		for (const auto& f : user_files) {
+			if (std::filesystem::is_directory(f)) {
+				for (const std::filesystem::directory_entry& ent :
+		             std::filesystem::recursive_directory_iterator(f)) {
+					std::string fl(ent.path().string());
+					if (fl.length() < 4 || ent.path().extension() != ".irt")
+						continue;
+					irt_files.push_back(fl);
+                }
```
weltling, line 229: nit: Looks like WS issue, closing brace with wrong indent. Thanks

ir-test.cxx
weltling: LGTM!

Makefile
weltling: Thanks for the further checks, @stkeke. I went double checking and seems indeed it's about the standard implementation integrity. Despite `C++17` is requested, seems the `multiline` flag is missing in some lower compiler version, even though it has some C++17 features.

It seems also, that some distro can backport features independently into an older compiler, so it might be hard to say if having a version `X` tells 100% some feature is absent.  Perhaps, it can be worked around with some other `std::regex` flags :/ For now, keeping it as simple as passing the `CXX=` arg seems appropriate.

Thanks!

ir-test.cxx
stkeke: I should have checked diff in the website. Let me check what happened in my editor.

Makefile
stkeke: @weltling agree!

stkeke:
@weltling @dstogov Thanks for the code review. All comments are resolved in code.

dstogov:
Looks good. I approved github actions for your PR and it seems something is wrong on Windows.
See https://github.com/dstogov/ir/actions/runs/4740374326/jobs/8417007111?pr=24
   
After Windows build is fixed, can you please also add the following Makefile changes and rebase the PR. 

```diff
diff --git a/Makefile b/Makefile
index f1a02be..d6dfaab 100644
--- a/Makefile
+++ b/Makefile
@@ -91,10 +91,10 @@ test: $(BUILD_DIR)/ir $(BUILD_DIR)/ir-test
 	$(BUILD_DIR)/ir $(SRC_DIR)/test.ir --dump --save 2>$(BUILD_DIR)/test.log
 	$(BUILD_DIR)/ir $(SRC_DIR)/test.ir --dot $(BUILD_DIR)/ir.dot
 	dot -Tpdf $(BUILD_DIR)/ir.dot -o $(BUILD_DIR)/ir.pdf
-	BUILD_DIR=$(BUILD_DIR) SRC_DIR=$(SRC_DIR) $(BUILD_DIR)/ir-test
+	$(BUILD_DIR)/ir-test $(SRC_DIR)/tests
 
 test-ci: $(BUILD_DIR)/ir $(BUILD_DIR)/ir-test
-	BUILD_DIR=$(BUILD_DIR) SRC_DIR=$(SRC_DIR) $(BUILD_DIR)/ir-test --show-diff
+	$(BUILD_DIR)/ir-test --show-diff $(SRC_DIR)/tests
 
 clean:
 	rm -rf $(BUILD_DIR)/ir $(BUILD_DIR)/ir_test $(BUILD_DIR)/*.o \
diff --git a/win32/Makefile b/win32/Makefile
index 37fef36..0b8b23f 100644
--- a/win32/Makefile
+++ b/win32/Makefile
@@ -118,15 +118,11 @@ $(BUILD_DIR)\ir-test.exe: $(SRC_DIR)/ir-test.cxx
 test: $(BUILD_DIR)\ir.exe $(BUILD_DIR)\ir-test.exe
 	set PATH=$(VCPKG_DIR)\installed\$(VCPKG_TRIPLET)\bin:%%PATH%%
 	$(BUILD_DIR)\ir.exe $(SRC_DIR)\test.ir --dump --save $(BUILD_DIR)\test.log
-	set BUILD_DIR=$(BUILD_DIR)
-	set SRC_DIR=$(SRC_DIR)
-	$(BUILD_DIR)\ir-test.exe
+	$(BUILD_DIR)\ir-test.exe $(SRC_DIR)\tests
 
 test-ci: $(BUILD_DIR)\ir.exe $(BUILD_DIR)\ir-test.exe
 	set PATH=$(VCPKG_DIR)\installed\$(VCPKG_TRIPLET)\bin:%%PATH%%
-	set BUILD_DIR=$(BUILD_DIR)
-	set SRC_DIR=$(SRC_DIR)
-	$(BUILD_DIR)\ir-test.exe --show-diff
+	$(BUILD_DIR)\ir-test.exe --show-diff $(SRC_DIR)\tests
 
 clean:
 	del /f /q $(BUILD_DIR)\*.obj $(BUILD_DIR)\*.exe $(BUILD_DIR)\*.pdb $(BUILD_DIR)\*.ilk $(BUILD_DIR)\*.h
```

stkeke:
@dstogov Sure. let me check what is wrong with Windows. Will update Makefile after fixing.

PR #27, opened 2023-05-03
Add ir.a target in Makefile

tekknolagi:
This makes it easier to use as a library:

```
  $ make ir.a
  ...
  $ gcc test.c ir.a -o test
  $ ./test
  :)
  $
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,9 @@ $(BUILD_DIR):
 $(BUILD_DIR)/ir: $(OBJS_COMMON) $(OBJS_IR)
 	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS) -lcapstone
 
+$(BUILD_DIR)/ir.a: $(OBJS_COMMON)
+	ar r $@ $^
+
 $(BUILD_DIR)/ir_test: $(OBJS_COMMON) $(OBJS_IR_TEST)
 	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS) -lcapstone
 


dstogov (changes requested):
@tekknolagi thanks for the proposal.

We didn't think about distribution this like a library yet. Currently we use IR in a new generation JIT for PHP and embed the necessary IR source files through the PHP build system.

I agree, the library may be helpful for others.
I think, we may accept this, if we start using ``ir.a`` to build ``ir`` and ``ir_test``.
Similar thing should be done for Windows (may be not in the same PR).
Than ``Makefile`` may be extended with ``install`` and ``uninstall`` targets.

CC @weltling

tekknolagi:
I may look into this -- thanks

Makefile
```diff
@@ -54,6 +54,9 @@ $(BUILD_DIR):
 $(BUILD_DIR)/ir: $(OBJS_COMMON) $(OBJS_IR)
 	$(CC) $(CFLAGS) -o $@ $^ $(LDFLAGS) -lcapstone
 
+$(BUILD_DIR)/ir.a: $(OBJS_COMMON)
```
weltling, line 57: Usually the name should be prefixed with `lib`, eg. `libir.a`.

thanks

weltling:
> I agree, the library may be helpful for others.
> I think, we may accept this, if we start using `ir.a` to build `ir` and `ir_test`.

Ditto. Switching to depend on the `.a` target instead ob `.o` targets should be done in the same PR, so then still both types of artifacts are validated.

Btw, this still wouldn't exclude offering a way of drop-in sources, but i guess we can also see to package a source archive suitable for that, or maybe have a doc on how to embed sources. 

> Than `Makefile` may be extended with `install` and `uninstall` targets.

Looking at this, one could still say for these purposes having something like a dedicated `cmake` defs would serve better? While the `Makefile` is simple and quick for the development, delivering libraries and headers, plus handling dependencies, would be still better served by something more powerful. In the end, we might still end up producing dynamic libs as well, split debug info, etc.

Thanks

dstogov:
Merged as a part of https://github.com/dstogov/ir/pull/75

tekknolagi:
Cool!

PR #28, opened 2023-05-05
Fix linkage to libdl

jserv:
This patch resolves undefined reference to symbols in libdl.

Error message before this:
```
/usr/bin/ld: ir_emit.o: in function `ir_resolve_sym_name':
ir_emit.c:249: undefined reference to `dlsym'
/usr/bin/ld: ir_disasm.o: in function `ir_disasm_resolver':
ir_disasm.c:302: undefined reference to `dladdr'
collect2: error: ld returned 1 exit status
```


weltling (approved):
LGTM

PR #29, opened 2023-05-05
Fix build with older libstdc++

jserv:
The current 'test' target fails with older versions of libstdc++ found in various Linux distributions. The error appears to be related to the lack of support for regex::multiline in these older libraries.

Error messages:
```
g++ -O3 -std=c++17 ir-test.cxx -o ir-test
ir-test.cxx: In member function 'void ir::test::parse()':
ir-test.cxx:33:59: error: 'multiline' is not a member of std::__cxx11::regex
 33 | # define REGEX_FLAGS std::regex::ECMAScript | std::regex::multiline
    |                                                           ^~~~~~~~~
ir-test.cxx:126:61: note: in expansion of macro 'REGEX_FLAGS'
126 |                         std::regex sect_reg("^--[A-Z]+--$", REGEX_FLAGS);
    |                                                             ^~~~~~~~~~~
```

Tested on Ubuntu Linux 22.04-LTS.

### DIFF
--- a/ir-test.cxx
+++ b/ir-test.cxx
@@ -30,7 +30,7 @@
 #else
 # define PATH_SEP "/"
 # define EXE_SUF ""
-# define REGEX_FLAGS std::regex::ECMAScript | std::regex::multiline
+# define REGEX_FLAGS std::regex::ECMAScript
 # define DIFF_CMD "diff"
 # define DIFF_ARGS "--strip-trailing-cr -u"
 # define IR_ARGS ""


weltling:
Thanks for the PR. This seems to have been discussed recently https://github.com/dstogov/ir/pull/24#discussion_r1169413489 and it seems that there's indeed no good way to ensure the multiline flags are present even though C++17 is requirement. My only concern is, also as per the previous discussion, that without this flag things won't work properly. So fixing the compilation alone doesn't seem a proper solution if the parsing won't work properly.

In the previous discussion, we also talked about using g++-12 explicitly, Would you try that please?

Thanks

weltling:
Btw which gcc version is in question?

Thanks

jserv:
> Btw which gcc version is in question?

 have attempted to use gcc-9, gcc-10, and gcc-11, but each of them results in the same build breakage.

weltling (changes requested):
Unfortunately without the multiline flag the test parsing won't work. Perhaps one could check if an older compiler has some other flag or combination of flags with a similar effect.

Also it looks like the exit status is propagated not correctly, as the test run is still green. 

Thansk

PR #30, opened 2023-05-07
Improve cross-compilation

jserv:
This commit introduces support for the CROSS_COMPILE variable, which is commonly used in cross-compilation projects such as the Linux kernel.

When CROSS_COMPILE is set, the build process will use it to detect the proper architecture for code generation. If CROSS_COMPILE is not set, the build system will assume a native build by consulting the results of "uname" command. The SYSROOT variable is used along with CROSS_COMPILE to prepend both header and library default directories.

In addition, users can enforce gcc-multilib to generate 32-bit code under x86-64 platforms by specifying TARGET=x86.

Examples:
1. Arm64 code generation make CROSS_COMPILE=aarch64-none-linux-gnu-
2. Arm64 code generation with sysroot make CROSS_COMPILE=aarch64-none-linux-gnu- SYSROOT=$(HOME)/php/ARM64
3. 32-bit x86 code generation make TARGET=x86

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,8 @@
-# TRAGET may be "x86_64" or "x86" or "aarch64"
-TARGET     = x86_64
 # BUILD can be "debug" or "release"
 BUILD      = debug
 BUILD_DIR  = .
 SRC_DIR    = .
 
-CC         = gcc
-CXX        = g++
 BUILD_CC   = gcc
 CFLAGS     = -Wall -Wextra -Wno-unused-parameter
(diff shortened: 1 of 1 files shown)


weltling (changes requested):
If this scenario is going to be supported, the change should be tested. It would be thinkable to add a parallel set of runs to the CI matrix, that would exercise all three Linux build scenarios that are currently present. For the `aarch64` variant, QEMU could be also used to run the tests.

Thanks

Issue #23, opened 2023-04-11
doc: Add examples directory


weltling:
The idea behind this is to create a collection of a code snippets that can be used for both education purposes and later for embedding into documentation.

PR #37, opened 2023-05-21
doc: Add examples folder

weltling:
Laying the base for adding examples. A couple of examples are added and more should be to go. Some better common structure for examples is also to be worked out.

Fixes: #23

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,8 @@ TARGET     = x86_64
 BUILD      = debug
 BUILD_DIR  = .
 SRC_DIR    = .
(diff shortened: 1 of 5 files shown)


stkeke:
@weltling Thanks for the examples. This is exactly what I am looking for ... and I am also writing some simple example programs to learn and demonstrate IR construction APIs usage.

weltling:
Thanks for checking, @stkeke. Great minds think alike :) We can sure converse on this as both the code and the doc. I'd be intended to add far more examples and expand on the comments in the follow up additions.

Thanks

PR #39, opened 2023-05-23
A simple framework for writing IR examples

stkeke:
1) exmplfrm.h:  default run_myfunc and default main() function
2) exmpltmpl.c: template for writing example program
3) README.md:   simple introduction and how to write examples

Benefits and Usage:
1) Easy to change main() to add functions like dump dot file, emit C
   code, change optimization level, or emit assembly code for all examples
2) Default run_myfunc() provides a standard way to run parameter-less IR
   function.
3) Custom run_myfunc() offers a customary way to run IR function with
   parameters.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,9 @@ OBJS_COMMON = $(BUILD_DIR)/ir.o $(BUILD_DIR)/ir_strtab.o $(BUILD_DIR)/ir_cfg.o \
 	$(BUILD_DIR)/ir_cpuinfo.o
 OBJS_IR = $(BUILD_DIR)/ir_main.o
 OBJS_IR_TEST = $(BUILD_DIR)/ir_test.o
+
+#FIXME(Tony): Will become terribly long when we have many many examples.
+#             Consider to use examples/Makefile of their own.
 EXAMPLE_EXES = $(EXAMPLES_BUILD_DIR)/0001-basic $(EXAMPLES_BUILD_DIR)/0001-while $(EXAMPLES_BUILD_DIR)/0005-basic-runner-func
 
 all: $(BUILD_DIR) $(BUILD_DIR)/ir $(BUILD_DIR)/ir_test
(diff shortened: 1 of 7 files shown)


stkeke:
@dstogov @weltling Please review and advise. This example framework is meant to simplify example writing and let us focus on IR function programming and running.

dstogov:
1. I think for IR newbies it's going to be easier to see the whole program in a single C file (without template includes).
2. Examples may demonstrate something more complex than generation and calling of a single function. (e.g. linking between several functions, disassembling/debugging, custom compilation pipelines, etc). Templates won't help.

stkeke:
@dstogov Yes, agree. This is also what I am concerned to make things complex with template. 
Actually, with ir_jit_compile(), the source code already becomes reasonably short.

Makefile
```diff
@@ -47,6 +47,9 @@ OBJS_COMMON = $(BUILD_DIR)/ir.o $(BUILD_DIR)/ir_strtab.o $(BUILD_DIR)/ir_cfg.o \
 	$(BUILD_DIR)/ir_cpuinfo.o
 OBJS_IR = $(BUILD_DIR)/ir_main.o
 OBJS_IR_TEST = $(BUILD_DIR)/ir_test.o
+
+#FIXME(Tony): Will become terribly long when we have many many examples.
+#             Consider to use examples/Makefile of their own.
```
weltling, line 52: Ditto. Guess this can be a separate improvement once more examples are filled. Similarly will be also to see on how to reorganize examples by topic, etc.

Thanks

weltling (commented):
@stkeke I'd be on the same page with @dstogov regarding this topic - for an example clarity is crucial. This means it'd be a straightforward uncompressed code explaining one matter per example, with clarifying comments if needed, but with the focus on readability and assessability in first place. A code duplication is here not an issue. IMO a framework would defeat the purpose of explaining things effectively by adding an extra layer of unnecessary abstraction 

Thanks

stkeke:
Agree. Let's keep thing simple so far and drop this PR.

PR #43, opened 2023-06-10
msvc: capstone: Move to using upstream instead of vcpkg

weltling:
Moving to the usage of the upstream sources removes layers of complexity present in VCPKG and allows for more flexibility with regard to the dependency handling. Things done:

- Switch to building capstone based from an upstream tag
- Explicitly remove any unused arch support in the build
- Use static lib to simplify on-prem usage
- Cache the capstone artifacts in the pipeline
- Automatically reset cache based on the values in Makefile and VC++ toolset version

The current capstone version is still kept 5.0-rc2. While evaluating a possible downgrade to 4.0.2 as used by the Linux CI, some test regressions are observed on the x64 side. If needed, those might be spent a closer look, otherwise keeping the version seems ok.

### DIFF
--- a/.github/workflows/push.yml
+++ b/.github/workflows/push.yml
@@ -88,23 +88,36 @@ jobs:
     strategy:
       matrix:
         arch: [x86, amd64]
-        include:
-          - vcpkg_dir: .\win32\vcpkg
-          - vcpkg_triplet: x86-windows
-            arch: x86
-          - vcpkg_triplet: x64-windows
-            arch: amd64
     runs-on: windows-2022
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


dstogov:
Great! The build is green again. Thank you very much.

It may make sense to separate everything capstone related into a separate Makefile. (e.g. win32/third-party/capstone/Makefile) and call nested make.

weltling:
Thanks for the quick check. Indeed, with the nested makefile, it'll make it cleaner. I'll plan to evaluate further on that.

Thanks

dstogov:
It looks like something is broken again
See https://github.com/dstogov/ir/actions/runs/5270236470/jobs/9529458175

weltling:
Huh, weird it is. I'll plan to do another pass on that. At the first glance, VC++ toolset version has changed in the failing run. 

Thanks

PR #49, opened 2023-09-22
Test Unix x86 with asan/ubsan in CI

iluuu1994:


Makefile
```diff
@@ -10,32 +10,32 @@ EXAMPLES_BUILD_DIR = $(BUILD_DIR)/examples
...
+override CFLAGS += -Wall -Wextra -Wno-unused-parameter
 LDFLAGS    = -lm -ldl
 PHP        = php
 LLK        = llk
 #LLK        = $(PHP) $(HOME)/php/llk/llk.php
 
 ifeq (debug, $(BUILD))
- CFLAGS += -O0 -g -DIR_DEBUG=1
+ override CFLAGS += -O0 -g -DIR_DEBUG=1
 endif
```
dstogov, line 21: I've never known about the "override" before.
I think, it may be better to use a different BUILD type. e.g. "asan" in addition to "release" and "debug".

iluuu1994: https://www.gnu.org/software/make/manual/html_node/Override-Directive.html

> If a variable has been set with a command argument (see [Overriding Variables](https://www.gnu.org/software/make/manual/html_node/Overriding.html)), then ordinary assignments in the makefile are ignored. If you want to set the variable in the makefile even though it was set with a command argument, you can use an override directive, which is a line that looks like this:
>
> override variable = value

I think `override` is a good flag either way, because otherwise users can still set `CFLAGS` via `make CFLAGS=...` but local `CFLAGS` in the Makefile are completely ignored. This can lead to unexpected behavior. I.e. in this case it will cancel the `-Wall -Wextra -Wno-unused-parameter`, and `-O2 -g` with the `release` build, etc. This is not immediately obvious when the removal of the flags doesn't break the build.

dstogov: OK

PR #56, opened 2023-11-30
Fix unused var warning on x86-32

iluuu1994:
See https://github.com/php/php-src/actions/runs/7040744160/job/19162218677


dstogov (commented):
I already noticed this.
I'm currently work on a big patch, and this change is going to be committed as a part of it.

dstogov:
Merged as part of https://github.com/dstogov/ir/commit/76e6418cae873f7ff794e0dba4de0e0c2c7fcf21

PR #61, opened 2024-01-01
Fixup examples by making ir_init initialize the ret_type as well

danog:
`ret_type` always needs to be initialized to -1 for all return builders to work, not sure why it is done separately (immediately after ir_init) in php-src, this moves the init inside of ir_init, thus fixing the examples.

### DIFF
--- a/examples/mandelbrot.c
+++ b/examples/mandelbrot.c
@@ -192,7 +192,7 @@ int main(int argc, char **argv)
 	ir_save(&ctx, stderr);
 	ir_dump_live_ranges(&ctx, stderr);
 	f = fopen("mandelbrot.dot", "w+");
-	ir_dump_dot(&ctx, f);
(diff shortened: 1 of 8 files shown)


examples/mandelbrot.c
```diff
@@ -192,7 +192,7 @@ int main(int argc, char **argv)
 	ir_save(&ctx, stderr);
 	ir_dump_live_ranges(&ctx, stderr);
 	f = fopen("mandelbrot.dot", "w+");
-	ir_dump_dot(&ctx, f);
+	ir_dump_dot(&ctx, "dump", f);
```
dstogov, line 195: This part is merged via f3a2717dc2d193d4a0780cd822796ed6e1611ddc

dstogov:
Examples are fixed in a bit different way via 7835df5831e24cd57628570f11d6c0d4a9593f2f. Thanks.

PR #62, opened 2024-01-13
easing freebsd build approach.

devnexen:


Makefile
```diff
@@ -47,6 +48,13 @@ ifeq (aarch64, $(TARGET))
   DASM_ARCH  = aarch64
   DASM_FLAGS = -M
 endif
+ifeq (freebsd, $(OS))
+  CC=clang
+  BUILD_CC=clang
+  override CFLAGS += -I/usr/local/include
+  override BUILD_CFLAGS += -I/usr/local/include
+  LDFLAGS += -L/usr/local/lib -lutil
```
dstogov, line 56: why do you need ``/usr/local/include``?

dstogov: Do you know what symbols are imported from  ``util``?

Makefile
```diff
@@ -47,6 +48,13 @@ ifeq (aarch64, $(TARGET))
   DASM_ARCH  = aarch64
   DASM_FLAGS = -M
 endif
+ifeq (freebsd, $(OS))
+  CC=clang
+  BUILD_CC=clang
```
dstogov, line 53: May be we should use ``cc``?

Makefile
```diff
@@ -1,5 +1,6 @@
-# TRAGET may be "x86_64" or "x86" or "aarch64"
+# TARGET may be "x86_64" or "x86" or "aarch64"
 TARGET     = x86_64
+OS         = linux
```
dstogov, line 3: I think we may use ``$(shell uname -s)`` for OS detection.

Makefile
devnexen: > why do you need `/usr/local/include`?

capstone

devnexen: > Do you know what symbols are imported from `util`?

kinfo_getproc

dstogov: Can we avoid this dependency using ``sysctl``? Like in example at https://man.freebsd.org/cgi/man.cgi?query=sysctl&sektion=3&apropos=0&manpath=FreeBSD+10.0-RELEASE

dstogov: > capstone (in FreeBSD non system packages are in /usr/local)

Did you build capstone yourself?

devnexen: Not at all, FreeBSD package system put all in /usr/local

devnexen: > https://man.freebsd.org/cgi/man.cgi?query=sysctl&sektion=3&apropos=0&manpath=FreeBSD+10.0-RELEASE

should be possible I ll have a look later in the day

dstogov (approved):
Looks fine. Thanks.

PR #64, opened 2024-01-17
netbsd support fix.

devnexen:
tests fails because of mapping permissions changes. they need to be pre authorized beforehand with PROT_MPROTECT macro.

while at it, easing netbsd build itself, unlike freebsd no fake libdl,
 capstone in /usr/pkg.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,14 @@ ifeq (FreeBSD, $(OS))
   LDFLAGS += -L/usr/local/lib
 endif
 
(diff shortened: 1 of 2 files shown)


devnexen:
@dstogov seems you re in between major changes, previous commit fail for those too :)

dstogov:
> @dstogov seems you re in between major changes, previous commit fail for those too :)

yeah, the tests failures are unrelated. This is my bug.

PR #78, opened 2024-03-17
Fix IR JIT build with -Werror=strict-prototypes

petk:
This fixes warning when using -Wstrict-prototypes or -Werror=strict-prototypes.


dstogov:
Hi, thank for catching this, but it's better to use the proper prototype (int, char**)
Fixed via 5f82e7eeb848b58d17707650fb7a25d1e4634f5b

petk:
I see, of course, that main is the program entry. Thanks for the fix.

PR #79, opened 2024-03-17
Fix LLVM constant string escaping emit

nielsdos:


dstogov:
Ohhh. Shame on me :|
Thank you for catching this.

PR #80, opened 2024-03-22
Extend GCM with an ability to split partially-dead nodes

dstogov:
The code is based on the Benedikt Meurer's idea first implemented in V8. See: https://codereview.chromium.org/899433005

### DIFF
--- a/ir.h
+++ b/ir.h
@@ -531,8 +531,9 @@ void ir_strtab_free(ir_strtab *strtab);
 
 /* debug related */
 #ifdef IR_DEBUG
-# define IR_DEBUG_SCCP         (1<<27)
-# define IR_DEBUG_GCM          (1<<28)
+# define IR_DEBUG_SCCP         (1<<26)
+# define IR_DEBUG_GCM          (1<<27)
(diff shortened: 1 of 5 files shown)


dstogov:
This increases the speed of ``minilua`` compiled with ``IR`` almost twice. (0.091 sec instead of 0.180)
 
```sh
clang -Wno-everything -O -Xclang -disable-llvm-passes -w -c -emit-llvm -o tmp.bc dynasm/minilua.c
opt tmp.bc --passes='function(mem2reg)' -S -o minilua.ll
ir --llvm-asm minilua.ll --run ./bench/mandelbrot.lua
```

dstogov:
hi @cliffclick

I'm very interested in getting your opinion about this GCM extension (and may be some related ideas).
I'm not the author of the algorithm, and it was never described in papers.

cliffclick:
Sounds reasonable.  I wrestled with the notion, but had too many other fish to fry at the time.  Seemed low payoff, but more or less free.  Also, the X86 machine ops would clone "simple" ops for free into the larger X86 addressing mode, so again some simple/cheap cases got handled already.  Never did the general case.

PR #81, opened 2024-03-26
Implement e-SSA PI nodes

arnaud-lb:
This implements e-SSA PI constraint nodes, without range propagation.

This uses the folding engine to replace comparison ops into constants when possible, and in turn the SCCP engine can elide IF/SWITCH/GUARD. I only implemented folding for `EQ` right now, but I will add other ops. I also did not add support for symbolic ranges.

PI node signature has the form `PI(def, RANGE(min, max))` (op2 is a RANGE op). It should be possible to change this to `PI(def, min, max)`, but the folding engine supports only two operands currently.

PI nodes must be added manually, like PHI nodes. They are replaced by COPY ops after SCCP so that other passes do not have to be aware of them.

### DIFF
--- a/ir.c
+++ b/ir.c
@@ -1399,6 +1399,29 @@ bool ir_list_contains(const ir_list *l, ir_ref val)
 	return 0;
 }
 
+bool ir_remove_pis(ir_ctx *ctx)
+{
+	ir_ref n, i;
+	ir_insn *insn;
+
+	for (i = IR_UNUSED + 1, insn = ctx->ir_base + i; i < ctx->insns_count;) {
+		n = insn->inputs_count;
+
+		if (insn->op == IR_PI) {
+			ir_use_list_remove_all(ctx, insn->op2, i);
+			insn->op2 = IR_UNUSED;
+			insn->op = IR_COPY;
+			insn->inputs_count = 1;
+		}
+
+		n = ir_insn_inputs_to_len(n);
+		i += n;
+		insn += n;
+	}
+
+	return true;
+}
+
 static uint32_t ir_hashtab_hash_size(uint32_t size)
 {
 	size -= 1;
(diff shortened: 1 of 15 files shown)


ir.c
```diff
@@ -1399,6 +1399,29 @@ bool ir_list_contains(const ir_list *l, ir_ref val)
...
+	ir_insn *insn;
+
+	for (i = IR_UNUSED + 1, insn = ctx->ir_base + i; i < ctx->insns_count;) {
+		n = insn->inputs_count;
+
+		if (insn->op == IR_PI) {
+			ir_use_list_remove_all(ctx, insn->op2, i);
+			insn->op2 = IR_UNUSED;
+			insn->op = IR_COPY;
+			insn->inputs_count = 1;
```
dstogov, line 1414: From the first look, I don't see a big reason in converting ``PI`` to ``COPY``.

``COPY`` is not going to be removed, and have to be handled by all the following passes. Actually, we may handle ``PI`` in the same way as ``COPY`` and keep the range info. On the other hand this will may additional overhead for all the following passes.

It's also possible to remove ``PI`` completely (turning it into NOP and re-connecting) input/def/use chains. T

dstogov: This may be done at the second phase of SCCP, that removes constant calculations and redundant copies.

I don't know what approach is better yet.

dstogov (commented):
Look. I'm not ready for this.
I don't have a complete picture in my head yet (how e-SSA and ranges should be implemented and handled) .
This PR is a start of a big research work and nobody can estimate the profit and cost of the future solution.
I'm appreciate your initial attempt and  would be glad  if you could try to continue this research. 

As I sorter term sub-task, I would recommend to think about the constant GUARD elimination.

PR #84, opened 2024-04-03
Fix some minor typos in README

javiereguiluz:
I was reading the changes you made in https://github.com/dstogov/ir/commit/352306899eead546fb1c650331d0894f032060b3 (by the way, congrats on the amazing work making compilation time 40x faster than `gcc -o2`) and I found some minor typos in the README file, so I propose some fixes in this PR. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,13 @@
 # IR - Lightweight JIT Compilation Framework
 
 IR Framework is a practical solution for implementing JIT in medium-size projects. 
-It defines Intermediate Representaion (IR), provides a simple API for IR construction and
+It defines Intermediate Representation (IR), provides a simple API for IR construction and
 a set of algorithms for optimization, scheduling, register allocation and code
 generation. The resulting generated in-memory code, may be directly executed.
 
 This is not a stable finished product yet. It’s still under active development.
 It was started as a base for development of the next generation JIT compiler for PHP-9,
-but it's completly PHP independent.
+but it's completely PHP independent.
 
 A presentation about IR framework design and implementation details is available at
 [researchgate](https://www.researchgate.net/publication/374470404_IR_JIT_Framework_a_base_for_the_next_generation_JIT_for_PHP).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,7 +30,7 @@ in our IR, we don’t have any variables, their versions and name. Everything
 is represented by computation Nodes and Edges between them. Of course, we
 have special PHI Node that represents the Phi() function.
 
-Despite, our IR is graph based, internally, it’s represented as a plain
+Internally, our IR is graph-based and represented as a plain
```
dstogov, line 33: This changes the meaning of the original sentence. Graph based IR usually assume independently allocated nodes with direct pointers between them. Usage of plain arrays for that is an uncommon decision.
May be this should be **Internally, our graph-based IR is represented as a plain array**

README.md
```diff
@@ -70,9 +70,9 @@ LuaJIT [3].
 
 This pass implements a classical algorithm originally designed by M. N. Wegman
 and F. K. Zadeck [4] for SSA form. Unification of data and control dependencies
-made its implementation even simple. Despite of constant propagation itself
+made its implementation even simpler. Despite of constant propagation itself
 this pass also performs global Copy Propagation and re-applies the folding rules.
-At the end all the “dead” instructions (instructions that result are not used)
+At the end all the “dead” instructions (instructions that are not used)
```
dstogov, line 75: We can't check if the instruction is used or not. We may check if the value produced by the instruction (its result) is used. may be: **which results**?

dstogov (changes requested):
Thank you very much!
I'm ashamed :(

Could you please correct sentences in the two commented places.

README.md
javiereguiluz: I'm sorry! I've changed this as you proposed.

dstogov (approved):

README.md
dstogov: > Sorry for this one too. I reworded to keep the original meaning.

No problems and thank you.

PR #87, opened 2024-06-01
Add support for AArch64 FreeBSD dynamic TLS

nielsdos:
This change is necessary to support https://github.com/php/php-src/pull/11236 for IR JIT. Currently that PR is only for 8.2-8.3 but I plan on adding IR JIT support once this is merged.


dstogov (approved):

dstogov:
Thanks. I've merged this into PHP master.

PR #90, opened 2024-06-10
Sync DISASM_INTEL_SYNTAX condition #if/ifdef/defined (-Wundef)

petk:
For this one I'm not sure what would be better here. To define it to 0 or to do the `#ifdef`. In PHP <= 8.3 JIT engine there was also this one used but disabled with value 0.

### DIFF
--- a/ir_disasm.c
+++ b/ir_disasm.c
@@ -29,6 +29,10 @@
 #include <capstone/capstone.h>
 #define HAVE_CAPSTONE_ITER
 
+#ifndef IR_DISASM_INTEL_SYNTAX
+# define IR_DISASM_INTEL_SYNTAX 0
+#endif
+
 typedef struct _ir_sym_node {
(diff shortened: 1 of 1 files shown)


ir_disasm.c
```diff
@@ -365,7 +365,7 @@ int ir_disasm(const char    *name,
 	}
 #  endif
 	cs_option(cs, CS_OPT_DETAIL, CS_OPT_ON);
-#  if DISASM_INTEL_SYNTAX
+#  ifdef DISASM_INTEL_SYNTAX
```
dstogov, line 368: It seems this code was copy-pasted from the old PHP JIT.
I think, It's better to define IR_DISASM_INTEL_SYNTAX at start.

```c
#ifned IR_DISASM_INTEL_SYNTAX
# define IR_DISASM_INTEL_SYNTAX 0
#endif
```

PR #93, opened 2024-10-04
Use external __jit_debug_descriptor when IR_EXTERNAL_GDB_ENTRY is defined

arnaud-lb:
If there are multiple `__jit_debug_descriptor` symbols, GDB may pick any, and `ir_gdb_register_code()` has no effect.

### DIFF
--- a/ir_gdb.c
+++ b/ir_gdb.c
@@ -500,13 +500,14 @@ typedef struct _ir_gdbjit_descriptor {
 	struct _ir_gdbjit_code_entry *first_entry;
 } ir_gdbjit_descriptor;
 
+#ifdef IR_EXTERNAL_GDB_ENTRY
+extern ir_gdbjit_descriptor __jit_debug_descriptor;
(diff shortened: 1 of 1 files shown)


dstogov:
Where do you see the problem? (PHP+linux_x86_64 seems to work fine).
Probably, you are right. I just like to know more about the problem.

arnaud-lb:
I've had the problem on the same environment (php + linux x86_64). The exact problem I'm having is that GDB is not aware of JIT symbols (so I don't see stubs in backtraces for example).

My understanding is that the main binary's definition should be given precedence by the linker (so `ir_gdb_register_code()` would use the `__jit_debug_descriptor` of the php binary), but the behavior I'm observing is that `ir_gdb_register_code()` uses the symbol of `opcache.so`. I'm not sure why. Maybe it's due to `dlopen()`.

Building with `IR_EXTERNAL_GDB_ENTRY` + this patch fixes the issue for me.

PR #95, opened 2024-11-24
Add support for MUSL AArch64 dynamic TLS

nielsdos:
This is the IR part of https://github.com/php/php-src/pull/16924.
Only difference with FreeBSD is that the dtv offset is -8 instead of 0.


dstogov (approved):
I can't test this anyway. I hope this is right.

nielsdos:
> I hope this is right.

I tested this and also maintainer of Alpine tested this in their CI.

PR #106, opened 2025-01-22
Propagate code_buffer when creating a new ir_ctx in ir_schedule

sgraham:
A trimmed ir_ctx is created during ir_schedule(), but this loses a user-provided code_buffer if one was in the original ir_ctx.

When code_buffer isn't provided, the many VirtualAlloc calls when compiling lots of small functions take ~70% of the total runtime in my compiler. (!)

### DIFF
--- a/ir_gcm.c
+++ b/ir_gcm.c
@@ -1328,11 +1328,13 @@ restart:
 	new_ctx.cfg_edges = ctx->cfg_edges;
 	ctx->cfg_blocks = NULL;
 	ctx->cfg_edges = NULL;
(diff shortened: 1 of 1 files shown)


dstogov (approved):
I didn't get this problem. because I didn't use ``ir_jit_compile()`` directly (it was introduced for example) and assigned ``ctx->code_buffer`` right before ``ir_emit_code()`` (see ir_main.c).

Anyway, this patch is right. Thanks!

sgraham:
Ah, maybe I should have my own set of compile stages.

I thought it was probably better to stick with ir_jit_compile() to make sure I was using an ordering that worked together properly.

PR #109, opened 2025-03-03
Fix php/php-src#17966: Symfony JIT 1205 assertion failure

nielsdos:
In this [example code](https://github.com/php/php-src/issues/17966),
`ir_split_partially_dead_node()` tries to create a clone of an IR_COPY instruction, but that instruction has an extra data op in op2. IR tries to add uses to op2 but it's not actually a real ir_ref. In this case, op2 is the number 1; if it were an ir_ref it would correspond to the IR_START node, which seems wrong. This patch adds an extra check to see if it's a real input.

I'm not sure this is fully right, or if this is the best way to fix it. Happy to take advice.

This was blocking me from making further measurements on https://github.com/php/php-src/pull/17909

### DIFF
--- a/ir_gcm.c
+++ b/ir_gcm.c
@@ -401,9 +401,10 @@ static bool ir_split_partially_dead_node(ir_ctx *ctx, ir_ref ref, uint32_t b)
 	for (i = 1; i < clones_count; i++) {
 		clones[i].ref = clone = ir_emit(ctx, insn->optx, insn->op1, insn->op2, insn->op3);
 		insn = &ctx->ir_base[ref];
-		if (insn->op1 > 0) ir_use_list_add(ctx, insn->op1, clone);
-		if (insn->op2 > 0) ir_use_list_add(ctx, insn->op2, clone);
-		if (insn->op3 > 0) ir_use_list_add(ctx, insn->op3, clone);
+		/* Depending on the flags in IR_OPS, these can be references or data. */
+		if (insn->op1 > 0 && insn->inputs_count >= 1) ir_use_list_add(ctx, insn->op1, clone);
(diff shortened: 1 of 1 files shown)


dstogov:
Thanks! The fix seems right. I'll think a bit more tomorrow.

dstogov:
Look, the following is independent from your fix:

``IR_COPY`` and especially ``IR_COPY(_, 1)`` (or ``HARD_COPY``) are not the first class citizens in sea-of-nodes IR.
That ``1`` in op2 disables removing of the ``COPY``, but in general all ``COPY`` should be removed.

We have to insert ``COPY`` in ``ir_bind()`` to make distinct between nodes bound to different PHP stack slots. (unfortunately, binding concept doesn't work in general, but this is a different story).

The other COPY-es in ``zend_jit_ir.c`` should be not necessary (some of them are "early" optimizations the other may be required for some bug workaround). I'll need to revisit this.

dstogov:
I thought ``IR_COPY`` is a single problematic instruction, but ``IR_PROTO`` may also cause the same problem.

nielsdos:
> Look, the following is independent from your fix: (...)

Right, I understand.

> The other COPY-es in zend_jit_ir.c should be not necessary (some of them are "early" optimizations the other may be required for some bug workaround). I'll need to revisit this.

I see, I did not know about this.

Is there anything extra you want me to do for this PR/fix? It sounds like you want to review the IR_COPY-es first?

dstogov:
> Is there anything extra you want me to do for this PR/fix? It sounds like you want to review the IR_COPY-es first?

Yes. I plan to do this tomorrow. I plan to commit this and merge into PHP on Thursday evening.

dstogov (approved):

PR #110, opened 2025-03-19
Fix condition when to use local variable in ir_sccp_remove_unfeasible_merge_inputs

nielsdos:
This seems like a typo.


dstogov (approved):
Thanks

PR #111, opened 2025-04-01
Optimize OR(x, PWR2) => BTS, AND(x, ~PWR2) => BTR

arnaud-lb:
Add rule to compile OR/AND instructions with large power of two op2 to a BTS/BTR instruction.

This avoids loading op2 into a register when it does not fit in an imm32, thus reducing instruction count and code size.

Before:

```asm
movabsq $9223372036854775808, %rcx ; 10 bytes
orq %rcx, %rax                     ;  3 bytes
```

After:

```asm
btsq $63, %rax                     ;  4 bytes
```

GCC does this optimization: https://godbolt.org/z/ThP3aEvEW (but not Clang)

### DIFF
--- a/ir_x86.dasc
+++ b/ir_x86.dasc
@@ -650,6 +650,26 @@ IR_ALWAYS_INLINE ir_mem IR_MEM(ir_reg base, int32_t offset, ir_reg index, int32_
 ||	}
 |.endmacro
 
+/* Like ASM_REG_IMM_OP, but op1 accepts r16,r32,r64 (not r8) */
+|.macro ASM_REG16_IMM_OP, op, type, op1, op2
+||	switch (ir_type_size[type]) {
+|| 		default:
+||			IR_ASSERT(0);
+||		case 1:
+||		case 2:
+|			op Rw(op1), (op2 & 0xffff)
+|| 			break;
+||		case 4:
+|			op Rd(op1), op2
+|| 			break;
+|.if X64
+||		case 8:
+|			op Rq(op1), op2
+|| 			break;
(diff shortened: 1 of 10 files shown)


arnaud-lb:
Test failure is because different capstone versions print an immediate value differently. (looks like that version 4 prints it in hex, and version 5 in decimal). I'm not sure what to do.

dstogov:
> Test failure is because different capstone versions print an immediate value differently. (looks like that version 4 prints it in hex, and version 5 in decimal). I'm not sure what to do.

I already saw something similar because of different capstone versions. I'll try to find a workaround.

arnaud-lb:
It's likely a bug in capstone. Here, capstone uses hex format when the `imm` is > HEX_THRESHOLD, but `imm` is signed, so large values are printed as decimal anyway.

I've updated the test to use a value without the high bit set (https://github.com/dstogov/ir/pull/111/commits/73bfc880a15c24c1e22786725c409a9eff378085), and it's all green now.

dstogov:
After landing this PR I see changes in the PHP-Parser code generated by function JIT.

```diff
-       orl $0x80000000, 0x28(%r15)
+       movl 0x28(%r15), %eax
+       btsl $0x1f, %eax
+       movl %eax, 0x28(%r15)
```

At first, BTS doesn't have any advantage over OR for immediate values that fit into 32-bit (In addition I remember, that Intel Optimization Manual recommended not to use BT* instructions at least on old CPUs). At second, at the diff above BTS misses the ability of address fusion.

Can you please take a look?

PR #112, opened 2025-04-08
Include generated files from the include path

arnaud-lb:
Using `#include "generated-file.h"` searches in the directory containing the file being compiled. When compiling out of the source dir, this may include the wrong file (e.g. a stalled, previously generated file in the source dir).

Using `#include <generated-file.h>` ensures that we search the file in the include path first, which may give priority to the build dir.


dstogov (approved):
Thanks!

PR #115, opened 2025-07-01
Prevent clobbering of TAILCALL op2 in epilogue

arnaud-lb:
TAILCALL op2 may be overridden when it is assigned to a preserved register, as the epilogue is emitted before the call:

```
{
	l_1 = START(l_4);
	uintptr_t f = PARAM(l_1, "p", 1);
	l_2 = CALL(l_1, f);
	l_3 = TAILCALL(l_2, f);
	l_4 = UNREACHABLE(l_3);
}
```

```
test:
	pushq %rbx
	movq %rdi, %rbx
	callq *%rbx
	popq %rbx
	jmpq *%rbx
```

This can not be trivially solved in RA, as the set of used preserved regs is not known yet.

Here I move op2 to a tmp reg in `ir_emit_tailcall()` when op2 is in `used_preserved_regs`.

### DIFF
--- a/ir_aarch64.dasc
+++ b/ir_aarch64.dasc
@@ -4943,6 +4943,26 @@ static void ir_emit_tailcall(ir_ctx *ctx, ir_ref def, ir_insn *insn)
 		return;
 	}
 
+	/* Move op2 to a tmp register before epilogue if it's in
+	 * used_preserved_regs, because it will be overridden. */
+
+	ir_reg op2_reg = IR_REG_NONE;
+	if (!IR_IS_CONST_REF(insn->op2)) {
+		op2_reg = ctx->regs[def][2];
+
+		IR_ASSERT(op2_reg != IR_REG_NONE);
+		if (IR_REGSET_IN((ir_regset)ctx->used_preserved_regs, IR_REG_NUM(op2_reg))) {
+			ir_reg orig_op2_reg = IR_REG_NUM(op2_reg);
+			op2_reg = IR_REG_INT_TMP;
+			if (IR_REG_SPILLED(orig_op2_reg)) {
+				ir_emit_load(ctx, IR_ADDR, op2_reg, insn->op2);
+			} else {
+				ir_type type = ctx->ir_base[insn->op2].type;
+				| ASM_REG_REG_OP mov, type, op2_reg, orig_op2_reg
+			}
+		}
+	}
+
 	ir_emit_epilogue(ctx);
 
 	if (IR_IS_CONST_REF(insn->op2)) {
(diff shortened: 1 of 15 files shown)


ir_aarch64.dasc
```diff
@@ -4955,13 +4975,8 @@ static void ir_emit_tailcall(ir_ctx *ctx, ir_ref def, ir_insn *insn)
 			|	br Rx(IR_REG_INT_TMP)
 		}
     } else {
-		ir_reg op2_reg = ctx->regs[def][2];
-
 		IR_ASSERT(op2_reg != IR_REG_NONE);
-		if (IR_REG_SPILLED(op2_reg)) {
-			op2_reg = IR_REG_NUM(op2_reg);
-			ir_emit_load(ctx, IR_ADDR, op2_reg, insn->op2);
-		}
```
dstogov, line 4964: Can we have a spilled ``op2_reg`` that not interference with  ``ctx->used_preserved_reg``? 
I think you lost this case.

ir_x86.dasc
```diff
@@ -9229,6 +9229,44 @@ static void ir_emit_tailcall(ir_ctx *ctx, ir_ref def, ir_insn *insn)
...
+				mem = ir_ref_spill_slot(ctx, insn->op2);
+			}
+			ir_reg base = IR_MEM_BASE(mem);
+			ir_reg index = IR_MEM_INDEX(mem);
+			if ((base != IR_REG_NONE && IR_REGSET_IN((ir_regset)ctx->used_preserved_regs, base)) ||
+					(index != IR_REG_NONE && IR_REGSET_IN((ir_regset)ctx->used_preserved_regs, index))) {
+				op2_reg = IR_REG_RAX;
+
+				ir_type type = ctx->ir_base[insn->op2].type;
+				| ASM_REG_MEM_OP, mov, type, op2_reg, mem
```
dstogov, line 9265: Use ``ir_emit_load_mem_int()``

tests/x86_64/tailcall_002.irt
```diff
@@ -0,0 +1,27 @@
...
+test:
+	pushq %rbx
+	subq $0x10, %rsp
+	movq %rdi, %rbx
+	movq %rbx, (%rsp)
+	callq *%rbx
+	movq %rbx, %rdi
+	addq $0x10, %rsp
+	popq %rbx
+	jmpq *(%rsp)
```
dstogov, line 26: The code looks wrong. ``%rsp`` is changed between ``movq %rbx, (%rsp)`` and ``jmpq *(%rsp)``

dstogov (changes requested):
Thanks for the attempt to fix this and for the new test cases.
I think, the idea should work, but I see few problems.
See comments...

We probably should also check for ``%rsp`` and ``%rbp`` in the fused memory load.

tests/x86_64/tailcall_002.irt
arnaud-lb: Indeed, I didn't notice that

ir_aarch64.dasc
arnaud-lb: Yes, I missed this case. I've updated the PR and added tests for that.

PR #122, opened 2026-01-10
Fix floating point guard jcc x86-64 codegen

ndossche:
Guard codegen relies on inverting the condition in
`ir_emit_guard_cmp_fp`. However, for floating point this results in
different behaviour w.r.t. NAN.

For example, if we have `GUARD(_, LT(_, _), _)` then we should only
continue executing code if `LT(_, _)` is satisfied.
Inverting the condition and branch targets makes this `GE(_, _)`, but
NANs don't trigger the `GE(_, _)` condition, resulting in continuing
executing the code if a NAN was present.

To solve this: track whether the condition was inverted and
conditionally generate jumps based on the parity flag.
Because of this change, the "swap operands to avoid P flag check"
optimization also is affected, so fix this as well.

Related to https://github.com/php/php-src/issues/20880

### DIFF
--- a/tests/x86_64/guard_001.irt
+++ b/tests/x86_64/guard_001.irt
@@ -5,10 +5,12 @@ x86_64
 --ARGS--
 -S
 --CODE--
+extern func @test(): void;
+func guard(double, double): int32_t
 {
 	int32_t c_1 = 0;
 	l_1 = START(l_50);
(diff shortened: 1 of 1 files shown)


dstogov:
Thanks for looking into the problem.
However, I made a more general fix. See https://github.com/dstogov/ir/pull/123

PR #135, opened 2026-04-05
ci: add job for macos-26 arm64

jorgsowa:


dstogov:
See https://github.com/dstogov/ir/pull/150 and https://github.com/dstogov/rcc/pull/9
Most IR tests failures are caused by "insignificant differences" in  disassembler output, but some by differences in ABI.

dstogov (approved):
Cool! This is exactly what I missed.
I'm merging this.

Please separate all the failed MAC/SYSV sensitive tests (except ./tests/debug.aarch64/tailcall_002.irt) as you started in https://github.com/dstogov/ir/pull/194

./tests/debug.aarch64/tailcall_002.irt for MAC should XFAIL similar to x86.

PR #141, opened 2026-04-07
Fix undef segv

dstogov:


weltling (approved):
Looks good. The fallback approach is cleaner than the per op field I used. 

This supersedes #137, I'll close it.

Thanks

PR #146, opened 2026-05-30
Fix README mandelbrot path typo

abnormal749:
## Summary

- Fix README commands to reference `bench/mandelbrot.ir`, matching the existing benchmark file.

## Validation

- `git diff --check HEAD~1 HEAD`


dstogov (approved):
Thanks!

PR #149, opened 2026-06-15
fuzz: Add structured graph fuzz harness and fix folding overflows it found

weltling:
This adds a second libFuzzer harness that fuzzes the IR builder and optimization pipeline directly, plus two folding fixes the harness found.

### Harness

The existing harness `fuzz_ir.c` feeds raw bytes to the text IR parser, so most mutated inputs are rejected before reaching the optimizer. The new harness `fuzz_graph.c` decodes each input blob into a structurally valid IR function instead, so inputs reach the optimizer and the code emitter. This is the core of the harness. Later features like loops and conditions will be built on top of it.

Decoding maps the leading bytes to a working type and parameter count, then turns each following group of bytes into one type preserving operation over earlier nodes, ending with a return of the last value. The build targets `fuzz-graph-O0`, `fuzz-graph-O1` and `fuzz-graph-O2` select the optimization level. The existing `fuzz_ir.c` is renamed to `fuzz_text.c` so the pair reads as text versus graph.

### Fixes found

- `MUL(C_U16, C_U16)`. Both `uint16_t` operands promote to `int`, so the product overflowed signed `int`. Multiply through `uint32_t`.
- `MUL/DIV(NEG, C_I)`. The constant was negated on a signed `int64_t`, which is undefined for `INT64_MIN`. Negate through `uint64_t`. The release result is unchanged since the negation wraps to itself.

Both were caught by the UBSan build of the harness, and the folded output is identical on a release build.


dstogov (commented):
The ``ir_fold.h`` fixes look right. Thanks!

According to fuzzer, it would be great to completely separate it, moving  everything related from ``Makefile`` to ``fuzz/Mafefile``. Is this difficult to do?

In case some day fuzzer is going to execute generated code, it should do this inside a container... This is not necessary right now.

weltling:
> The `ir_fold.h` fixes look right. Thanks!
> 
> According to fuzzer, it would be great to completely separate it, moving everything related from `Makefile` to `fuzz/Mafefile`. Is this difficult to do?
> 
> In case some day fuzzer is going to execute generated code, it should do this inside a container... This is not necessary right now.

Thanks for the review.

Yep, I'm going to restructure all the fuzz code under the subfolder then, which will make it cleaner and not interfere with the production code.

Thanks

weltling:
Implemented the suggested change. Everything `fuzz` is now encapsulated in the subfolder and rebuild on its own. All the other details from the PR description are still valid, too. A simple command sequence to check:

```bash
make -C fuzz
make -C fuzz corpus
./fuzz/build/fuzz-graph-O2
```

With this I think it's ready for a review now.

Thanks

dstogov:
Hi @weltling, I merged this and simplified ``fuzz/Makefile`` to build ``libir.a`` through the main ``Makefile``.
Please verify that everything works as expected.

weltling:
@dstogov, thanks for merging and for the Makefile simplification. I verified it on current master `bef94ef` - everything works as expected. As a follow up, adding control structures seems to make most of sense, I'll be getting to it next.

Thanks

PR #155, opened 2026-06-28
fuzz: Add control flow and loops to the structured graph fuzzer

weltling:
Adds control structures to the structured graph fuzzer. The graph builder can now emit an if and else diamond that merges through a PHI, and a simple natural loop driven by an induction PHI.

Also, the PR fixes a use after free in the iterative condition optimizer that the new control flow path uncovered. While simplifying an IF condition, for example IF(SHR(x, c)) into IF(AND(x, mask)), the optimizer allocates a constant that can grow the constant pool and reallocate ir_base, leaving cached instruction pointers dangling. The fix refetches those pointers after the allocation in the condition helper and in the IF and GUARD optimizers.

A new regression tests/folding/fold_015.irt reproduces the crash under the address sanitizer.


dstogov (approved):
The fix looks right.
Please, next time create different pull requests for fixes and fuzzer improvements.

weltling:
> The fix looks right. Please, next time create different pull requests for fixes and fuzzer improvements.

Thanks for merging, and noted on splitting fixes from fuzzer work. The use after free just rode along here since the new control flow path surfaced it.

Next on my fuzzer list: integer div and mod, float to int and any other type conversions. Separately, memory ops and indirect calls. Please let me know if you see more to be covered.

Thanks

PR #156, opened 2026-06-28
Fix dead loop guard that never re-pins OVERFLOW after code motion

iliaal:
The use-list scan in ir_gcm_schedule_late() that pulls an IR_OVERFLOW projection into its ADD/SUB/MUL_OV block runs `for (n = use_list->count; ...; n < 0; ...)`. n starts at a non-negative count, so the body never executes; every other use-list scan in the file uses n > 0. With the loop dead, an OV op relocated by GCM can leave its OVERFLOW projection in a different block, where it is emitted standalone (seto/setc) over already-clobbered flags. Using n > 0 restores the re-pin.


dstogov (approved):
Absolutely Right! Thanks!

PR #157, opened 2026-06-28
Fix stale ir_base pointer after ir_const() in the iterative optimizer

iliaal:
`ir_promote_d2f/f2d/i2i()` and the `ir_iter_opt()` driver cache an `ir_insn*` and store through it after a call that reaches `ir_const()`, which can grow the constant buffer (`ir_grow_bottom`), reallocate `ir_base`, and leave the cached pointer dangling. Re-fetch after each such call; the two-operand cases read both operand refs up front so the single re-fetch isn't preceded by a stale read. Deterministic ASAN repro, clean after the patch with no test regressions: `FP2FP(float)` over `ADD(double, FP2FP(double, param), const)` with the constant buffer at its limit gives a heap-use-after-free WRITE freed by `ir_grow_bottom` in `ir_promote_d2f`. The `ir_iter_optimize_condition/if/guard` re-fetches this PR originally also carried are now on master via #155, so the rebase drops them.

### DIFF
--- a/ir_sccp.c
+++ b/ir_sccp.c
@@ -1607,6 +1607,7 @@ static ir_ref ir_promote_d2f(ir_ctx *ctx, ir_ref ref, ir_ref use)
 {
 	ir_insn *insn = &ctx->ir_base[ref];
 	uint32_t count;
+	ir_ref op;
 
 	IR_ASSERT(insn->type == IR_DOUBLE);
 	if (IR_IS_CONST_REF(ref)) {
@@ -1642,7 +1643,9 @@ static ir_ref ir_promote_d2f(ir_ctx *ctx, ir_ref ref, ir_ref use)
 //				return ref;
 			case IR_NEG:
 			case IR_ABS:
-				insn->op1 = ir_promote_d2f(ctx, insn->op1, ref);
+				op = ir_promote_d2f(ctx, insn->op1, ref);
+				insn = &ctx->ir_base[ref];
+				insn->op1 = op;
 				insn->type = IR_FLOAT;
 				return ref;
 			case IR_ADD:
@@ -1652,10 +1655,17 @@ static ir_ref ir_promote_d2f(ir_ctx *ctx, ir_ref ref, ir_ref use)
 			case IR_MIN:
 			case IR_MAX:
 				if (insn->op1 == insn->op2) {
-					insn->op2 = insn->op1 = ir_promote_d2f(ctx, insn->op1, ref);
+					op = ir_promote_d2f(ctx, insn->op1, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op2 = insn->op1 = op;
 				} else {
-					insn->op1 = ir_promote_d2f(ctx, insn->op1, ref);
-					insn->op2 = ir_promote_d2f(ctx, insn->op2, ref);
+					ir_ref op1 = insn->op1;
+					ir_ref op2 = insn->op2;
+					op1 = ir_promote_d2f(ctx, op1, ref);
+					op2 = ir_promote_d2f(ctx, op2, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op1 = op1;
+					insn->op2 = op2;
 				}
 				insn->type = IR_FLOAT;
 				return ref;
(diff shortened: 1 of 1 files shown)


ir_sccp.c
```diff
@@ -1652,10 +1655,16 @@ static ir_ref ir_promote_d2f(ir_ctx *ctx, ir_ref ref, ir_ref use)
...
+					insn = &ctx->ir_base[ref];
+					insn->op2 = insn->op1 = op;
 				} else {
-					insn->op1 = ir_promote_d2f(ctx, insn->op1, ref);
-					insn->op2 = ir_promote_d2f(ctx, insn->op2, ref);
+					op = ir_promote_d2f(ctx, insn->op1, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op1 = op;
+					op = ir_promote_d2f(ctx, insn->op2, ref);
+					insn = &ctx->ir_base[ref];
```
dstogov: I think, it's better to reload ``insn`` once after both ``ir_promote_d2f()`` calls.

ir_sccp.c
```diff
@@ -1726,10 +1738,16 @@ static ir_ref ir_promote_f2d(ir_ctx *ctx, ir_ref ref, ir_ref use)
...
+					insn->op2 = insn->op1 = op;
 				} else {
-					insn->op1 = ir_promote_f2d(ctx, insn->op1, ref);
-					insn->op2 = ir_promote_f2d(ctx, insn->op2, ref);
+					op = ir_promote_f2d(ctx, insn->op1, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op1 = op;
+					op = ir_promote_f2d(ctx, insn->op2, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op2 = op;
```
dstogov: The same, please, reload ``insn`` once.

ir_sccp.c
```diff
@@ -1893,10 +1913,16 @@ static ir_ref ir_promote_i2i(ir_ctx *ctx, ir_type type, ir_ref ref, ir_ref use)
...
+					insn->op2 = insn->op1 = op;
 				} else {
-					insn->op1 = ir_promote_i2i(ctx, type, insn->op1, ref);
-					insn->op2 = ir_promote_i2i(ctx, type, insn->op2, ref);
+					op = ir_promote_i2i(ctx, type, insn->op1, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op1 = op;
+					op = ir_promote_i2i(ctx, type, insn->op2, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op2 = op;
```
dstogov: The same.

ir_sccp.c
```diff
@@ -1908,10 +1934,16 @@ static ir_ref ir_promote_i2i(ir_ctx *ctx, ir_type type, ir_ref ref, ir_ref use)
...
+					insn->op3 = insn->op2 = op;
 				} else {
-					insn->op2 = ir_promote_i2i(ctx, type, insn->op2, ref);
-					insn->op3 = ir_promote_i2i(ctx, type, insn->op3, ref);
+					op = ir_promote_i2i(ctx, type, insn->op2, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op2 = op;
+					op = ir_promote_i2i(ctx, type, insn->op3, ref);
+					insn = &ctx->ir_base[ref];
+					insn->op3 = op;
```
dstogov: The same.

dstogov (commented):
The fixes are right. Thanks!
Some of these fixes are already merged.
Please make the minor adjustments (see comments).

iliaal:
Done in all four `else` branches. They read both operand refs up front, so the second promote doesn't read the sibling operand through a stale `insn`. Also dropped a now-unused `p` left by the PHI-loop change.

iliaal:
Rebased onto master and dropped the `ir_iter_optimize_condition/if/guard` re-fetches, those landed via #155. What's left is the promote functions and the `ir_iter_opt` driver. Also collapsed the two-operand `else` branches to a single reload as requested.

dstogov (approved):
Thanks!
Now I see that your previous approach also might make sense.
I'll merge this and then try which one is better.

PR #158, opened 2026-06-28
Harden debug-only paths (IR verifier, GDB JIT builder) against out-of-bounds access

iliaal:
Two debug-only out-of-bounds hardening fixes found while auditing the framework.

ir_check(): after flagging an out-of-range instruction operand it still dereferenced ir_base[use], an out-of-bounds read on the input it just rejected; it now skips to the next operand. The constant-operand range check compared a negative const ref against consts_count and never fired; it now tests -use against consts_count to match the insn-ref check.

The GDB JIT ELF builder (ir_gdbjit_strz/uleb128/sleb128 and raw ctx->p writes) fills a fixed space[4096] stack buffer guarded only by an IR_ASSERT on the final size, which runs after the writes and is compiled out in release, so a long symbol or file name overflowed the stack. It now asserts ctx->p stays within space[] at each write site. A runtime bound wouldn't help: with an over-long name strz fills the buffer and the next raw DU16 write in debuginfo overflows instead (those fixed-width writes can't truncate), so the bound only relocates an overflow that real names never reach.

Both paths are debug-only (IR_DEBUG verifier and GDB JIT registration) and not reachable from normal codegen. No test regressions.

### DIFF
--- a/ir_gdb.c
+++ b/ir_gdb.c
@@ -198,9 +198,12 @@ static uint32_t ir_gdbjit_strz(ir_gdbjit_ctx *ctx, const char *str)
 {
 	uint8_t *p = ctx->p;
 	uint32_t ofs = (uint32_t)(p - ctx->startp);
-	do {
-		*p++ = (uint8_t)*str;
-	} while (*str++);
+	while (*str) {
+		IR_ASSERT(p < ctx->obj.space + sizeof(ctx->obj.space));
+		*p++ = (uint8_t)*str++;
+	}
+	IR_ASSERT(p < ctx->obj.space + sizeof(ctx->obj.space));
+	*p++ = '\0';
 	ctx->p = p;
 	return ofs;
 }
@@ -209,6 +212,7 @@ static uint32_t ir_gdbjit_strz(ir_gdbjit_ctx *ctx, const char *str)
 static void ir_gdbjit_uleb128(ir_gdbjit_ctx *ctx, uint32_t v)
 {
 	uint8_t *p = ctx->p;
+	IR_ASSERT(p + 5 <= ctx->obj.space + sizeof(ctx->obj.space));
 	for (; v >= 0x80; v >>= 7)
 		*p++ = (uint8_t)((v & 0x7f) | 0x80);
 	*p++ = (uint8_t)v;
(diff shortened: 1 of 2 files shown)


dstogov (changes requested):
- ``ir_check()`` fixes are completely right.
- ir_gdb.c fix looks incomplete. 
  - it silently ignores the tail of the string. I think it's better to do ``IR_ASSERT()``.
  - the same protection should be inserted into ``ir_gdbjit_uleb128()``, ``ir_gdbjit_sleb128()`` and others where ``ctx->p`` is dereferenced.

iliaal:
Done. `strz` asserts now instead of truncating, and the same bounds assert is in `uleb128`, `sleb128`, the raw `ctx->p` writes in `secthdr`/`symtab`, and `initsect_done` as a backstop for the DWARF macro writes.

I dropped the truncation rather than keeping it: bounding `strz` doesn't make release safe. With a name past 4096 it just fills the buffer and the next raw `DU16` in `debuginfo` overflows instead (confirmed under ASAN), and those fixed-width writes can't truncate. GDB JIT registration is debug-only and real names are far short of 4096, so the assert is the right guard; a bound would only relocate an overflow that can't happen in practice.

ir_gdb.c
```diff
@@ -212,6 +215,7 @@ static void ir_gdbjit_uleb128(ir_gdbjit_ctx *ctx, uint32_t v)
 	for (; v >= 0x80; v >>= 7)
 		*p++ = (uint8_t)((v & 0x7f) | 0x80);
 	*p++ = (uint8_t)v;
+	IR_ASSERT(p <= ctx->obj.space + sizeof(ctx->obj.space));
```
dstogov: Could you please, move the ``IR_ASSERT()`` above the ``for`` loop.
This should trigger assertion before the possible buffer overflow.

dstogov (changes requested):
Please do the final minor tweaks.

ir_gdb.c
iliaal: Moved above the loop in both `uleb128` and `sleb128`, with `p + 5 <= end` (max LEB128 bytes for a 32-bit value) so it trips before a straddling write rather than after.

PR #161, opened 2026-07-05
ir: Extend narrow integer constants after folding

weltling:
Arithmetic fold rules compute on the narrow typed fields, so a result that overflows the type width keeps garbage in the upper bits. The folder then reads the full `val.u64` for comparison and for the divisor guards, so a value like 256 in a `U8` passes the zero check and divides by zero.

This was surfaced by the structured graph fuzzer from #160, which hit it in `DIV(C_U8, C_U8)`.

The fix canonicalizes at `ir_fold_const` so every narrow integer result is zero or sign extended to its type width, covering every fold rule at once.

### DIFF
--- a/ir.c
+++ b/ir.c
@@ -1074,12 +1074,27 @@ ir_fold_copy:
 		return IR_FOLD_DO_COPY;
 	}
 ir_fold_const:
-	if (!ctx->use_lists) {
-		return ir_const(ctx, val, IR_OPT_TYPE(opt));
-	} else {
-		ctx->fold_insn.opt = IR_OPT(IR_OPT_TYPE(opt), IR_OPT_TYPE(opt));
-		ctx->fold_insn.val.u64 = val.u64;
-		return IR_FOLD_DO_CONST;
+	{
+		ir_type type = IR_OPT_TYPE(opt);
+
+		/* Extend a narrow integer result to its type width so no
+		 * fold rule sees garbage in the upper bits. */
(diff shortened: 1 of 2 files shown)


ir.c
```diff
@@ -1074,12 +1074,27 @@ ir_ref ir_folding(ir_ctx *ctx, uint32_t opt, ir_ref op1, ir_ref op2, ir_ref op3,
...
+		/* Extend a narrow integer result to its type width so no
+		 * fold rule sees garbage in the upper bits. */
+		if (IR_IS_TYPE_INT(type) && ir_type_size[type] < 8) {
+			uint32_t shift = (8 - ir_type_size[type]) * 8;
+			if (IR_IS_TYPE_SIGNED(type)) {
+				val.i64 = (int64_t)(val.u64 << shift) >> shift;
+			} else {
+				val.u64 = (val.u64 << shift) >> shift;
+			}
+		}
```
dstogov, line 1089: You have found a huge bug! :)

This happens because of ``C`` semantic that always extend operands to ``int``.
I have to think how to fix this better...
- fix each place in r_fold.h
- keep this patch
- keep this patch, but reduce ``ir_type_size[type] < 8`` to ``ir_type_size[type] < 4``.

dstogov (approved):
Your fix is fine. Thanks!

ir.c
weltling: Thanks for merging.

Yep, `< 4` is the exact bound, whereby dropping the 4 byte case would change no result.

Guarding each rule in `ir_fold.h` individually would be a bigger refactoring.

PR #162, opened 2026-07-05
ir: Guard signed integer MOD against overflow

weltling:
The signed MOD fold guarded only a zero divisor, so `INT64_MIN % -1` fed the remainder operator an overflowing case, which is undefined. `DIV` already guards this pair, but `MOD` did not.

This was surfaced by the structured graph fuzzer from #160, which hit it in `MOD(C_I64, C_I64)`.

The fix mirrors the `DIV` guard in `MOD` and emits the node instead of folding the constant.

### DIFF
--- a/ir_fold.h
+++ b/ir_fold.h
@@ -706,13 +706,44 @@ IR_FOLD(MOD(C_ADDR, C_ADDR))
 }
 
 IR_FOLD(MOD(C_I8, C_I8))
+{
+	IR_ASSERT(IR_OPT_TYPE(opt) == op1_insn->type);
+	if (op2_insn->val.i64 == 0
+	 || (op2_insn->val.i64 == -1 && op1_insn->val.i8 == INT8_MIN)) {
+		/* division by zero or signed overflow */
+		IR_FOLD_EMIT;
+	}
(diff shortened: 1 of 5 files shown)


ir_fold.h
```diff
@@ -711,8 +711,9 @@ IR_FOLD(MOD(C_I32, C_I32))
 IR_FOLD(MOD(C_I64, C_I64))
 {
 	IR_ASSERT(IR_OPT_TYPE(opt) == op1_insn->type);
-	if (op2_insn->val.i64 == 0) {
-		/* division by zero */
+	if (op2_insn->val.i64 == 0
+	 || (op2_insn->val.i64 == -1 && op1_insn->val.i64 == INT64_MIN)) {
```
dstogov, line 745: I suppose, this woks only for I64, but not enough for I32, I16, I8.
Right?

weltling: Right, good catch! I've been too focused on the fuzz finding. To stay consistent with `DIV` and the backend the guard has to be per width. I'll split it and redo the test.

Thanks!

weltling: Updated, ready for the next round. Thanks

dstogov (approved):
Looks fine! Thanks!

PR #163, opened 2026-07-07
feat: implement REDUCE_ADD/MUL/MIN/MAX vector reduction opcodes

steven-studio:
Following up on our earlier discussion about ir_emit_c() support for vector types.

I spent some time verifying the existing vector lowering first: ADD (vector-typed
scalar op), SPLAT, EXTRACT, REPLACE, and vector values merged across an if/else
branch (PHI) all already lower to correct, gcc-compilable C. So the emit-c work
I originally offered to help with turned out to be mostly solid already.

What I did find missing was REDUCE — there was no corresponding opcode at the
IR level at all yet, so this PR adds it:

- New opcodes: `IR_REDUCE_ADD`, `IR_REDUCE_MUL`, `IR_REDUCE_MIN`, `IR_REDUCE_MAX`
  in `ir.h`, following the existing convention of one opcode per operation kind
  (like EQ/NE/LT/... and ADD_OV/SUB_OV/...) rather than a generic op + kind
  parameter. Each takes a single vector operand (`d1`, same shape as SPLAT but
  the reverse direction: vector -> scalar).
- Builder macros (`ir_REDUCE_ADD` etc.) in `ir_builder.h`.
- `ir_emit_c()` lowering in `ir_emit_c.c`: ADD/MUL unroll into a flat expression
  (`a[0]+a[1]+...+a[N-1]`), MIN/MAX unroll into a ternary comparison chain.

Known limitation: MIN/MAX use plain C comparison, which doesn't implement IEEE
754 NaN semantics for min/max. Happy to tighten this if it matters for your use
case — wanted to get the basic shape right first.

Verified with hand-written `.ir` test cases plus a small C harness using the
corresponding GCC vector_size ABI:
- REDUCE_ADD([1,2,3,4]) = 10 ✓
- REDUCE_MAX([3,7,1,5]) = 7 ✓

Happy to add SHUFFLE/CONVERT next if this direction looks right to you, or
adjust the opcode/codegen design if you had something different in mind.

### DIFF
--- a/ir.h
+++ b/ir.h
@@ -180,7 +180,7 @@ typedef enum _ir_type {
 #define IR_IS_TYPE_VECTOR(t)   (((t) & IR_VECTOR_MASK) != 0)
 
 #define IR_VECTOR_BASE_TYPE(t) ((t) & IR_BASE_TYPE_MASK)
-#define IR_VECTOR_LENGTH(t)    (1 << ((((t) & IR_VECTOR_MASK) >> 4) - 1))
+#define IR_VECTOR_LENGTH(t)    (1U << ((((t) & IR_VECTOR_MASK) >> 4) - 1))
 
 #ifdef IR_64
 # define IR_SIZE_T          IR_U64
--- a/ir.h
+++ b/ir.h
@@ -123,11 +123,15 @@ extern "C" {
 # define IR_X86_I64 0
 #endif
 
-/* IR Type flags (low 4 bits are used for type size) */
-#define IR_TYPE_SIGNED     (1<<4)
-#define IR_TYPE_UNSIGNED   (1<<5)
-#define IR_TYPE_FP         (1<<6)
-#define IR_TYPE_SPECIAL    (1<<7)
+#ifndef IR_SIMD
+# define IR_SIMD 1
+#endif
+
+/* IR Type flags */
+#define IR_TYPE_SIGNED     (1<<0)
+#define IR_TYPE_UNSIGNED   (1<<1)
+#define IR_TYPE_FP         (1<<2)
+#define IR_TYPE_SPECIAL    (1<<3)
 #define IR_TYPE_BOOL       (IR_TYPE_SPECIAL|IR_TYPE_UNSIGNED)
 #define IR_TYPE_ADDR       (IR_TYPE_SPECIAL|IR_TYPE_UNSIGNED)
 #define IR_TYPE_CHAR       (IR_TYPE_SPECIAL|IR_TYPE_SIGNED)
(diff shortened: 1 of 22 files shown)


ir.h
```diff
@@ -341,6 +341,10 @@ typedef enum _ir_type {
 	_(EXTRACT,      d2,   def, def, ___) /* get element of vector       */ \
 	_(REPLACE,      d3,   def, def, def) /* set element of vector       */ \
 	_(SPLAT,        d1,   def, ___, ___) /* set all elements of vector  */ \
+	_(REDUCE_ADD,   d1,   def, ___, ___) /* sum all elements of vector  */ \
+	_(REDUCE_MUL,   d1,   def, ___, ___) /* product of all elements     */ \
+	_(REDUCE_MIN,   d1,   def, ___, ___) /* min of all elements         */ \
+	_(REDUCE_MAX,   d1,   def, ___, ___) /* max of all elements         */ \
```
dstogov: I didn't decide yet, if it's better to introduce separate ``REDUCE`` instruction(s) or emulate them through "SHUFFLE", "EXTRACT" and already existing vector and scalar operations.
 
- x86 don't have separate reduction instructions
- GCC and CLANG seem don't have special built-ins for this
- I didn't find reduction instructions in WASM Spec (may be I'm wrong)

So why should we introduce them and then care about IR construction and deconstruction during code generation?

- On the other hand LLVM has [Vector Reduction Intrinsics](https://llvm.org/docs/LangRef.html#vector-reduction-intrinsics)

- in case we introduce ``REDUCE``, I would prefer a single instruction, where ``op2`` specifies the horizontal operation (ADD, MUL, MAX, MIN, AND, OR, XOR)
```
_(REDUCE,   d1X1,   def, num, ___)
```

dstogov (commented):
Thanks for the PR. Everything is clear and fine.
The questions are only abut design (see comments and give your opinion).

> Verified with hand-written .ir test cases plus a small C harness using the
corresponding GCC vector_size ABI:

RCC (simd branch) now can translate C with ``__attribute__((__vector_size(N)))`` to IR.

> Happy to add SHUFFLE/CONVERT next if this direction looks right to you, or
adjust the opcode/codegen design if you had something different in mind.

Most probably, I'll implement them myself, by the end of the week. I might leave ``ir_emit_c()`` implementation part for you.

My priority now is implementation of some (minimal but good enough) subset of SIMD extension (IR and x86 back-end) and landing it to the ``master``.

ir.h
steven-studio: Thanks for the feedback! I agree with your reasoning — went ahead and reworked this into a single REDUCE opcode, where op2 holds the ir_op selecting the horizontal operation:

```c
_(REDUCE,   d1X1,   def, num, ___) /* horizontal reduce; op2 = ir_op (IR_ADD/IR_MUL/IR_MIN/IR_MAX) */
```

Updated `ir_builder.h` to a single `ir_REDUCE(type, vec, op)` macro, and `ir_emit_c()` now dispatches on `insn->op2` instead of on four separate opcodes. Same lowering strategy as before (flat expression for ADD/MUL, ternary-chain comparison for MIN/MAX), just consolidated.

Re-verified with the same `.ir` test cases + GCC `vector_size` C harness: `REDUCE_ADD([1,2,3,4])=10`, `REDUCE_MAX([3,7,1,5])=7`, both still correct.

Force-pushed the update to `simd` (amended the same commit rather than adding a new one, since this was purely a design revision of the same feature). Let me know if this matches what you had in mind, or if you'd want a different set of allowed op2 values (e.g. reserving room for AND/OR/XOR reductions later).

dstogov:
Something is completely wrong with the ``diff``. It includes my whole ``simd`` branch.

dstogov:
May be this is because, I had to rebase ``simd`` branch and force pushed it.

steven-studio:
Rebased simd onto the latest 8709564 and resolved the conflicts (mainly overlapping changes in ir.h, ir_builder.h, and ir_emit_c.c from the vector-type work). Force-pushed the cleaned-up branch — the diff should now only show the 3 commits for this PR (Improve support for vector types, ir_emit_c() support for vector types, feat: implement REDUCE vector reduction opcode), and GitHub shows no conflicts with the base branch anymore.

ir_dump.c
```diff
@@ -555,53 +555,23 @@ void ir_dump_codegen(const ir_ctx *ctx, FILE *f)
...
+		} else if (insn->op == IR_LABEL) {
+			fprintf(f, "label(%s)", ir_get_str(ctx, insn->val.name));
+		} else if (insn->op == IR_FUNC_ADDR) {
+			fprintf(f, "func *");
+			ir_print_const(ctx, insn, f, true);
+			ir_print_proto(ctx, insn->proto, f);
+		} else {
+			ir_print_const(ctx, insn, f, true);
 		}
+		fprintf(f, ";\n");
```
dstogov: This hunk looks unrelated revert.

ir_emit_c.c
```diff
@@ -669,6 +669,65 @@ static void ir_emit_replace(ir_ctx *ctx, FILE *f, ir_ref def, ir_insn *insn)
...
+	fprintf(f, "[0];\n");
+	for (j = 1; j < n; j++) {
+		fprintf(f, "\td_%d = (", ctx->vregs[def]);
+		ir_emit_ref(ctx, f, insn->op1);
+		fprintf(f, "[%u] %s d_%d) ? ", j, is_max ? ">" : "<", ctx->vregs[def]);
+		ir_emit_ref(ctx, f, insn->op1);
+		fprintf(f, "[%u] : d_%d;\n", j, ctx->vregs[def]);
+	}
+}
+
```
dstogov, line 730: This function is not used.

dstogov (commented):
Sorry for a long delay.

At this moment I think - IR doesn't need the REDUCE instruction.
Its introduction would require its implementation for each target and vector types, that is not trivial for x86.

For WASM->IR->C you may emulate REDUCE on IR level (using simple loop with EXTRACT and scalar operation or unrolled loop with SHUFFLE and vector operations). Anyway the emitted C code is going to be almost the same.

ir.h
steven-studio: Agreed on dropping REDUCE from the IR level given the per-target codegen burden — makes sense to close this out rather than rebase onto the newer simd work. Thanks for the detailed feedback! I'll keep an eye on where the x86 SIMD backend lands and revisit REDUCE emulation (loop+EXTRACT or unrolled SHUFFLE) at the WASM→IR→C layer if it becomes relevant there.

PR #165, opened 2026-07-11
Fix ir_emit_c() crash on LOAD with unused result

jorgsowa:
ir_emit_load() called ir_emit_def_ref() unconditionally, which asserts ctx->vregs[def] is set. A LOAD with no data users still keeps vregs[def] at 0, so the assert fires. Guard it like `ir_emit_call()` already does for CALL.


dstogov (approved):
Absolutely right! Thanks!

PR #166, opened 2026-07-11
Fix ir_emit_c() crash on ALLOCA with unused result

jorgsowa:
Same class of bug as the LOAD fix: ir_emit_alloca() called ir_emit_def_ref() unconditionally, which asserts ctx->vregs[def] is set. ALLOCA is MEM-flagged, not DATA-flagged, so an alloca whose pointer has no data users keeps vregs[def] at 0.


dstogov:
Fixed. Thanks!

PR #172, opened 2026-07-18
fix ir_emit_c() dropping VSTORE assignments and VAR declarations

jorgsowa:
At -O0, IR_VAR locals never get a vreg, so the declaration pass - gated on vregs[i] - never declared them. Separately, `ir_emit_vstore()` skipped re-emitting a store on the assumption its value was already written directly into the var by `ir_emit_def_ref();` that shortcut only ever fired for binary/unary ops, so storing a constant or PARAM into a var silently dropped the assignment.

Declare `IR_VAR` locals unconditionally. Remove the direct-write shortcut instead of patching its guesswork, and have `ir_emit_vstore()` always emit the assignment itself.


dstogov (approved):
Looks right. Thanks!

PR #173, opened 2026-07-18
fix "untill" to "until" in various comments

LamentXU123:
Fix "untill" to "until" in various comments in ir_cfg.c

This is wrongly pushed to php/php-src directly in https://github.com/php/php-src/commit/5b79dc5514b1ddce1e86baba037ab74618c36cbd as I wasn't aware of this dependency. Thanks you @jorgsowa for pointing it out!


dstogov (approved):
Thanks!

PR #180, opened 2026-08-18
Support for AArch64 SIMD vector instructions

dstogov:


dynasm/dasm_arm64.lua
```diff
@@ -801,8 +801,8 @@ map_op = {
   ["strh_*"]  = "78000000DwL",
   ["ldrh_*"]  = "78400000DwL",
   ["ldrsh_*"] = "78c00000DwL|78800000DxL",
-  ["str_*"]   = "b8000000DwL|f8000000DxL|bc000000DsL|fc000000DdL",
-  ["ldr_*"]   = "18000000DwB|58000000DxB|1c000000DsB|5c000000DdB|b8400000DwL|f8400000DxL|bc400000DsL|fc400000DdL",
+  ["str_*"]   = "b8000000DwL|f8000000DxL|bc000000DsL|fc000000DdL|3c800000DqL",
+  ["ldr_*"]   = "18000000DwB|58000000DxB|1c000000DsB|5c000000DdB|9c000000DqB|b8400000DwL|f8400000DxL|bc400000DsL|fc400000DdL|3cc00000DqL",
```
dstogov, line 805: Hi @MikePall

I'm working on SIMD instructions support for my IR/RCC projects.
Unfortunately, DynAsm doesn't support SIMD instructions on AArch64, and I don't have forces to properly implement this.
So I decided to encode SIMD instructions  "manually" using  ``| .long``.

I just tried to add  support for 128-bit ``LDR`` and ``STR`` instructions.
This seems work for labels, but doesn't properly work for ``ldr Rq(dst), [Rx(base], #offset]``, because of incorrect offset ``scaling``. DynAsm uses for scale only 2-bits, so scale 16 treated as 0.

Can this be fixed?

Don't care if this leads to complex  or dangerous changes.
It's not a problem for me to use "manual" encoding.
In the future I plan to replace DynAsm in my projects anyway.

PR #183, opened 2026-08-22
Add ABI header support to tester

jorgsowa:
Test `vaarg_004_aarch64.ir` currently fails on macOS arm64. I thought about implementation that would allow skipping certain tests that are not designed for Apple ABI.

In general I would like to extend `TEST_TARGET` to include ABI as well, so the tests may be skippable.

If target contains only `aarch64` then it doesn't consider ABI. If it's specificed i.e. aarch64-darwin, it means it's run only on 
macOS arm64.

If such approach is good, I can extend the TEST_TARGET to include other ABI, like windows', etc.

failing test:
```
TEST: VA_ARG 004: va_arg() expanded on AArch64 [./tests/run/vaarg_004_aarch64.irt]sh: line 1: 98433 Segmentation fault: 11  ./ir ./tests/run/vaarg_004_aarch64.ir --run > ./tests/run/vaarg_004_aarch64.out 2>&1
--- ./tests/run/vaarg_004_aarch64.exp	2026-08-21 22:02:01
+++ ./tests/run/vaarg_004_aarch64.out	2026-08-21 22:02:01
@@ -1 +1,2 @@
-sum 45
+
+termsig = 11
FAIL: VA_ARG 004: va_arg() expanded on AArch64 [./tests/run/vaarg_004_aarch64.irt]
```

### DIFF
--- a/tests/run/vaarg_004_aarch64_sysv.irt
+++ b/tests/run/vaarg_004_aarch64_sysv.irt
@@ -1,7 +1,7 @@
 --TEST--
-VA_ARG 004: va_arg() expanded on AArch64
+VA_ARG 004: va_arg() expanded on AArch64 SysV
 --TARGET--
-aarch64
+aarch64-sysv
 --ARGS--
 --run
 --CODE--
--- a/tools/tester.c
+++ b/tools/tester.c
@@ -178,13 +178,31 @@ static test *parse_file(const char *filename, int id)
 	return t;
 }
 
+static int match_target_name(const char *selector, const char *actual)
+{
+	size_t selector_len;
+
+	if (!selector || !actual) {
+		return 0;
+	}
+	if (strcmp(selector, actual) == 0) {
+		return 1;
+	}
+	selector_len = strlen(selector);
+	return strncmp(selector, actual, selector_len) == 0 && actual[selector_len] == '-';
+}
+
 static int skip_test(test *t)
 {
 	if (target && t->target) {
 		if (t->target[0] == '!') {
-			return strcmp(t->target + 1, target) == 0;
+			if (match_target_name(t->target + 1, target)) {
+				return 1;
+			}
 		} else {
-			return strcmp(t->target, target) != 0;
+			if (!match_target_name(t->target, target)) {
+				return 1;
+			}
 		}
 	}
 	return 0;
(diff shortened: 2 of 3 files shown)


tests/run/vaarg_004_aarch64_sysv.irt
```diff
@@ -1,7 +1,7 @@
 --TEST--
-VA_ARG 004: va_arg() expanded on AArch64
+VA_ARG 004: va_arg() expanded on AArch64 SysV
 --TARGET--
-aarch64
+aarch64-sysv
```
dstogov, line 4: I agree. We need make distinct between different ABIs.
I see the similar need testing PR with RISCV back-end.

tools/tester.c
```diff
@@ -178,13 +178,31 @@ static test *parse_file(const char *filename, int id)
...
+	size_t selector_len;
+
+	if (!selector || !actual) {
+		return 0;
+	}
+	if (strcmp(selector, actual) == 0) {
+		return 1;
+	}
+	selector_len = strlen(selector);
+	return strncmp(selector, actual, selector_len) == 0 && actual[selector_len] == '-';
```
dstogov, line 192: The current "minimal" approach with "whole match" / "not match" is already not sufficient.
We need something smarter (at least target triple with wilbcard, or even smarter).

dstogov (commented):
I think, GCC's test-suite approach should covers the minimal needs - ``/* { dg-do compile { target { i?86-*-* x86_64-*-* } } } */``

Also, the ``tester`` program doesn't have any special meaning about ``--TARGET--``.
It might check any environment variables...

I don't see how the best solution should look like yet.
I'll think about this.

It's also possible to introduce several ``--EXPECT--`` sections for different targets.
 
Any related ideas are welcome.

tools/tester.c
dstogov: Oh, I missed, you check for the whole match or exact cpu match. Thit is a bit better, but I would prefer general matching over several patterns (similar to fnmatch()).

 I'm trying to implement this now...

dstogov:
See https://github.com/dstogov/ir/pull/185
Let me know if you see any problems

jorgsowa:
It's similar approach that I wanted to take.

PR #187, opened 2026-08-27
Fix sparse set allocation size overflow

crystarm:
Fixes an integer-overflow risk when calculating the allocation size for IR sparse sets.

Uses ir_mem_calloc(size, 2 * sizeof(*set->data)) so the allocator performs the checked multiplication, while preserving the existing sparse/dense memory layout.

The issue was detected by static analysis: [BAD_ALLOC_ARITHMETIC ir_private.h:[500:14].log](https://github.com/user-attachments/files/31533111/BAD_ALLOC_ARITHMETIC.ir_private.h.500.14.log)

### DIFF
--- a/ir_private.h
+++ b/ir_private.h
@@ -507,9 +507,12 @@ typedef struct _ir_sparse_set {
 
 IR_ALWAYS_INLINE void ir_sparse_set_init(ir_sparse_set *set, uint32_t size)
 {
+	size_t alloc_size = (size_t)size * 2 * sizeof(*set->data);
+
 	set->size = size;
 	set->len = 0;
-	set->data = (uint32_t*)ir_mem_malloc(sizeof(uint32_t) * 2 * size) + size;
+	IR_ASSERT(!size || alloc_size / size == 2 * sizeof(*set->data));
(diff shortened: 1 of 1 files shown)


ir_private.h
```diff
@@ -509,7 +509,7 @@ IR_ALWAYS_INLINE void ir_sparse_set_init(ir_sparse_set *set, uint32_t size)
 {
 	set->size = size;
 	set->len = 0;
-	set->data = (uint32_t*)ir_mem_malloc(sizeof(uint32_t) * 2 * size) + size;
+	set->data = (uint32_t*)ir_mem_calloc(size, 2 * sizeof(*set->data)) + size;
```
dstogov: 1) ``calloc()`` fills the allocated memory by zero, that is not necessary and makes overhead.
2) In case of overflow ``calloc()`` will return NULL that should lead to following crash. 

A better solution, would be just checking for overflow (similar to ``zend_safe_address()``) and assertion.

dstogov (changes requested):

dstogov (approved):
This looks fine.

PR #188, opened 2026-08-28
fix: constant callee c emitter

jorgsowa:
Fix C emission for calls through constant function addresses.

Previously, constant callees were assumed to be named IR_FUNC references. When the callee was a raw function address, the generated C did not cast the address to the function's actual prototype before calling it.

Emit constant function addresses as typed function pointers using the call prototype, including the return type, parameter types, and variadic arguments.

### DIFF
--- a/ir_emit_c.c
+++ b/ir_emit_c.c
@@ -779,9 +779,34 @@ static void ir_emit_switch(ir_ctx *ctx, FILE *f, uint32_t b, ir_ref def, ir_insn
 	fprintf(f, "\t}\n");
 }
 
+static void ir_emit_func_ptr_cast(FILE *f, const ir_proto_t *proto, ir_type ret_type)
+{
+	uint32_t i;
+
+	ir_emit_c_type_name(proto ? proto->ret_type : ret_type, f);
+	fprintf(f, " (*)(");
+	if (proto && proto->params_count) {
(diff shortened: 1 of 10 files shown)


dstogov (changes requested):
This looks right, but a bit incomplete.
- you missed support for calling conventions
- may be it makes sense to use common code for  ``ir_emit_func_ptr_cast()`` and ``ir_emit_c_func_decl()`` (I'm not sure).
- ``ir_emit_c.c`` also misses correct support for IR_PROTO insn. It shouldn't be a big problem to fix that casting as well.

```
{
	uintptr_t c_1 = 0;
	bool c_2 = 0;
	bool c_3 = 1;
	l_1 = START(l_6);
	uintptr_t d_2 = PARAM(l_1, "p", 1);
	int32_t d_3 = PARAM(l_1, "a", 2);
	uintptr_t d_4 = PROTO(d_2, func (int32_t): int32_t __fastcall);
	int32_t d_5, l_5 = CALL/1(l_1, d_4, d_3);
	l_6 = RETURN(l_5, d_5);
}
```

``ir_emit_llvm.c`` has the same problem. Probably, it's better to fix it in next PR.

dstogov (approved):
Thanks!
I'll merge this and make minor clean-up (mainly renaming) on top of this.

PR #193, opened 2026-09-08
add IR_TLS_ADDR operation to get a __thread local variables address

### LINKED ISSUE #192: Feature request: ir_TLS_ADDR
I'm moving EG/CG into __thread storage so because IR can currently only load a tls addr I load a *self reference. If IR could expose the functionality of getting a tls vars address without loading it, that would both simplify code and improve performance.

henderkes:
WIP

closes https://github.com/dstogov/ir/issues/192

### DIFF
--- a/ir_x86.dasc
+++ b/ir_x86.dasc
@@ -13992,14 +13992,34 @@ static void ir_emit_lea(ir_ctx *ctx, ir_ref def, ir_type type)
 	}
 }
 
+|.macro TLS_LOAD_OR_ADDR, reg, offset, addr
+||	if (addr) {
+||		if (offset) {
+|			lea reg, aword [reg+offset]
+||		}
+||	} else {
+|		mov reg, aword [reg+offset]
+||	}
+|.endmacro
+
+|.macro STATIC_TLS_LOAD_OR_ADDR, reg, offset, addr
+||	if (addr) {
+|		mov reg, aword [0]
+|		TLS_LOAD_OR_ADDR reg, offset, addr
+||	} else {
+|		mov reg, aword [offset]
+||	}
+|.endmacro
+
 static void ir_emit_tls(ir_ctx *ctx, ir_ref def, ir_insn *insn)
 {
 	ir_backend_data *data = ctx->data;
 	dasm_State **Dst = &data->dasm_state;
 	ir_reg reg = IR_REG_NUM(ctx->regs[def][0]);
+	bool addr = insn->op == IR_TLS_ADDR;
 
 	if (ctx->use_lists[def].count == 1) {
-		/* dead load */
+		/* dead result */
 		return;
 	}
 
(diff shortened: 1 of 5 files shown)


ir_x86.dasc
```diff
@@ -13992,52 +13992,72 @@ static void ir_emit_lea(ir_ctx *ctx, ir_ref def, ir_type type)
 	}
 }
 
+|.macro TLS_LOAD_OR_ADDR, reg, offset, addr
+||	if (addr) {
+||		if (offset) {
+|			lea reg, aword [reg+offset]
```
dstogov, line 13998: this is the same as ``add reg, offset``

ir_x86.dasc
```diff
@@ -13992,52 +13992,72 @@ static void ir_emit_lea(ir_ctx *ctx, ir_ref def, ir_type type)
...
+||		}
+||	} else {
+|		mov reg, aword [reg+offset]
+||	}
+|.endmacro
+
+|.macro STATIC_TLS_LOAD_OR_ADDR, reg, offset, addr
+||	if (addr) {
+|		mov reg, aword [0]
```
dstogov, line 14007: Why do you relay on offset 0?
In TLS variant 1 TCB with offset 0 contains a pointer to DTV, that is not necessary a part of TCB.
You can't access "static" elements in TCB, using DTV base.

May be I'm wrong

dstogov (commented):
I agree that the existing implementation (TLS) is not good enough.

But I think that the proposed (TLS_ADDR)  should to be designed better.

Imagine the following code:
```c
extern __thread int x;
extern __thread int y;
void test(void) {
	y = x;
}
```

For static-local TLS the proposed implementation would require 6 instructions
```asm
	movl	%eax, %fs[0]
	leal	%eax, [%eax, x@TPOFF]
	movl	%eax, [%eax] 
	movl	%edx, %fs[0]
	leal	%edx, [%edx, y@TPOFF]
	movl	[%edx], %eax 
```

instead of required 2
```asm
	movl	%fs:x@TPOFF, %eax
	movl	%eax, %fs:y@TPOFF
```

henderkes:
> I agree that the existing implementation (TLS) is not good enough.
> 
> But I think that the proposed (TLS_ADDR) should to be designed better.
> 
> Imagine the following code:
> 
> ```c
> extern __thread int x;
> extern __thread int y;
> void test(void) {
> 	y = x;
> }
> ```
> 
> For static-local TLS the proposed implementation would require 6 instructions
> 
> ```assembly
> 	movl	%eax, %fs[0]
> 	leal	%eax, [%eax, x@TPOFF]
> 	movl	%eax, [%eax] 
> 	movl	%edx, %fs[0]
> 	leal	%edx, [%edx, y@TPOFF]
> 	movl	[%edx], %eax 
> ```
> 
> instead of required 2
> 
> ```assembly
> 	movl	%fs:x@TPOFF, %eax
> 	movl	%eax, %fs:y@TPOFF
> ```

Yes, the implementation is currently nothing more than a very inefficient copy of ir_TLS and not ready for review yet. I've only created tge draft to help illustrate what I'm after.

I wanted to continue working on this, but am currently busy with preparations for api con. I should come back for this end of next week but will need a while to read up on what guarantees are available to reach a single instruction access.

dstogov:
Anyway, you pointed to a real problem. I'll try to take care. Thanks!

PR #194, opened 2026-09-09
Add Darwin AArch64 varargs test

jorgsowa:
Test `args_001.irt` fails on Darwin. 
Example: https://github.com/dstogov/ir/actions/runs/29170932167/job/86753268624?pr=135#step:5:1131

With triple keyword in target, we can now specify target for ABI. Is this approach acceptable? There are more such tests failing on Darwin. I can migrate them properly in next PR.


dstogov (commented):
This makes sense, but unfortunately we have another serious blocker.
A lot of tests failed, because capstone on MacOS uses aliases (e.g. ``mov`` instead of ``movz``). See https://github.com/dstogov/ir/actions/runs/29170932167/job/86753268624?pr=135#step:5:583

It seems support for instruction aliases was extended in capstone v5 (but the ability to disable this at run-time will be available only in v6). I think, switching to capstone v4 should fix this. I don't know how to do this on Mac. Could you pleas try.

Fixes for specific tests are welcome after the blocker fix.

jorgsowa:
Closed in favor of https://github.com/dstogov/ir/pull/195

PR #195, opened 2026-09-10
tests: add macOS AArch64 call expectations

jorgsowa:
Creates tests for Darwin ABI and limits the existing aarch64 tests to sysv.

Supersedes https://github.com/dstogov/ir/pull/194


dstogov (approved):
Everything right. Thanks!
I was wrong about ``tailcall_002.irt``. It doesn't use variable arguments and not affected by differences in calling-convention.

PR #199, opened 2026-09-13
Fix aarch64 constant shift immediate range

jorgsowa:
Fix AArch64 code generation for constant shifts with shift counts outside the immediate range.


tests/aarch64/shift_const_sar32_32_011.irt
```diff
@@ -0,0 +1,17 @@
+--TEST--
+011: aarch64 constant SAR32 by 32 keeps immediates in range
+--TARGET--
+aarch64
+--ARGS--
+-O1 -S
```
dstogov: Again, It's better to create ``run`` target-independent tests and verify similar problems in x86 back-end as well. I wouldn't like to create too many separate tests. Better to combine few checks for specific problem in a single test.

ir_aarch64.dasc
```diff
@@ -3368,6 +3370,25 @@ static void ir_emit_shift_const(ir_ctx *ctx, ir_ref def, ir_insn *insn)
...
+		}
+		goto store;
+	}
+	if (type_size < 32 && shift >= type_size) {
+		if (insn->op == IR_SAR) {
+			shift = type_size - 1;
+		} else {
+			|	mov Rw(def_reg), wzr
+			goto store;
+		}
```
dstogov, line 3389: This need to be covered by ``run`` tests.
This is "C Undefined Behavior", I would prefer to make the result to be the same as GCC and/or CLANG (ideally the same for x86 and Aarch64, but not necessary).

dstogov (changes requested):

tests/run/sar_003.irt
```diff
@@ -0,0 +1,25 @@
...
+}
+func main(): int32_t
+{
+	uintptr_t c_1 = func test(int32_t): int32_t;
+	int32_t c_2 = -2147483647;
+	l_1 = START(l_3);
+	int32_t d_2, l_2 = CALL/1(l_1, c_1, c_2);
+	l_3 = RETURN(l_2, d_2);
+}
+--EXPECT--
```
dstogov, line 29: I wonder, do tests relay on exit code?

Of course, it would be better to print result(s) of SAR and add EXPECT-ed value(s).
This would also allow to reduce the number of tests, reusing the same ``test`` function for different values.

jorgsowa: Yes, they expect 0 exit code.

I have few ideas how test robustness may be improved. I will try to submit PRs soon

dstogov:
Closed in favour of https://github.com/dstogov/ir/commit/5c536d032c4da153475431783b245f827a630b0e

PR #200, opened 2026-09-14
Re-design instruction for TLS access

dstogov:
Now TLS_ADDR(module, offset) is used to get address of TLS variable. If "module" is negative then "static" TLS-address returned (TCB + offset), otherwize "module" is used as offset in Dynamic Thread Vector to return  *(TCB->DTV + module) + offset.

ir_TLS() macro is kept for compatibility.


ir_aarch64.dasc
```diff
@@ -6509,50 +6556,97 @@ static void ir_emit_guard_overflow(ir_ctx *ctx, ir_ref def, ir_insn *insn)
...
-		return;
-	}
 
 ||#ifdef __APPLE__
 ||	code = 0xd53bd060 | reg; // TODO: hard-coded: mrs reg, tpidrro_el0
 |	.long code
 |	and Rx(reg), Rx(reg), #0xfffffffffffffff8
-|//???	MEM_ACCESS_64_WITH_UOFFSET_64 ldr, Rx(reg), Rx(reg), #insn->op2, TMP1
-|//???	MEM_ACCESS_64_WITH_UOFFSET_64 ldr, Rx(reg), Rx(reg), #insn->op3, TMP1
+|//???	MEM_ACCESS_64_WITH_UOFFSET_64 ldr, Rx(reg), Rx(reg), #mod, TMP1
```
dstogov, line 6629: [jorgsowa](https://github.com/jorgsowa) could you please try to fix TLS access on aarch64-darwin.
I can't remember where did I get ``and Rx(reg), Rx(reg), #0xfffffffffffffff8`` from...

henderkes:
Thank you so much for getting on it. I can only give a small review for now, but I'll try to submit the tests for output verification I've already worked on tomorrow.

dstogov:
> Thank you so much for getting on it. I can only give a small review for now, but I'll try to submit the tests for output verification I've already worked on tomorrow.

I've tested only PHP-8.4/Linux/64-bit/ZTS (everything looks fine)
I hope the new design should be suitable for your EG()/CG() PR.

ir_aarch64.dasc
jorgsowa: I can test. FYI: I didn't get notification because you copied only url of my profile. You forgot @jorgsowa to notify me.

ir_aarch64.dasc
```diff
@@ -6509,50 +6556,97 @@ static void ir_emit_guard_overflow(ir_ctx *ctx, ir_ref def, ir_insn *insn)
...
+	if (insn->op2) {
+		|	add Rx(reg), Rx(reg), #insn->op2
+	}
+
+	if (IR_REG_SPILLED(ctx->regs[def][0])) {
+		ir_emit_store(ctx, IR_ADDR, def, reg);
+	}
+}
+
+static void ir_emit_tls_load(ir_ctx *ctx, ir_ref def, ir_insn *insn)
```
henderkes, line 6666: have you accounted for different spilling in the existing tls paths? Previously only pointer type loads had to be accounted for, but this could now load e.g. a floating point value, right?

henderkes:
I think you still need to take care of folding tls load operands directly into arithmetic operations. I'm not going to use that in php *right now*, but might later.

E.g. `x+load(tls_addr())` could emit

```asm
addl %fs:n, %eax
```
 instead of 

```asm
movq %fs:0, %rcx
addq $n, %rcx
addl (%rcx), %eax
```

dstogov:
> I think you still need to take care of folding tls load operands directly into arithmetic operations. I'm not going to use that in php _right now_, but might later.

Not "folding" but "fusion". Anyway, this optimization is possible without API change and could work out of the box.
May be I'll try it later.

henderkes:
I tickled the slop machine to adapt the slightly changed api for my existing php branch and got:

The PR fails validation:
- Direct TLS loads/stores emit single instructions, but comparisons still construct an address first.
- PHP already uses TLS arithmetic in EG(vm_stack_end) - EG(vm_stack_top). That triggers a folding assertion.
- Function JIT hits an entry-point assertion; numeric TLS offsets can also crash compilation.

henderkes:
https://github.com/henderkes/php-src/tree/testing/ir_tls_addr

ir_aarch64.dasc
jorgsowa: I created two regression tests for problems on Apple. Run output: https://github.com/dstogov/ir/actions/runs/34944799054/job/104301541747?pr=201

dstogov: > I created two regression tests for problems on Apple. Run output: https://github.com/dstogov/ir/actions/runs/34944799054/job/104301541747?pr=201

Thanks! This should be fixed (force pushed).

However, the main problem, that the generated code for TLS access on Mac/Aarch64 most probably incorrect (you see ``//???``). Currently, the implementation doesn't support "dynamic" TLS through DTV. Static TLS might be right ot not.

You may try to see what GCC/CLANG generates, and how ``__tls_get_addr()`` is implemented.

jorgsowa: For a native `_Thread_local` variable it generates a Mach-O TLV access:

```asm
adrp x0, _tls_var@TLVPPAGE
ldr  x0, [x0, _tls_var@TLVPPAGEOFF]  ; TLV descriptor
ldr  x8, [x0]
blr  x8
```

I need to dig deeper into this topic yet as I don't fully understand how it should work

ir_aarch64.dasc
```diff
@@ -6569,50 +6616,97 @@ static void ir_emit_guard_overflow(ir_ctx *ctx, ir_ref def, ir_insn *insn)
...
+	}
+
+	IR_ASSERT(def_reg != IR_REG_NONE && reg != IR_REG_NONE);
+	IR_ASSERT(addr_insn->op == IR_TLS_ADDR);
+	ir_emit_tls_base_addr(ctx, reg, addr_insn->op1);
+	mem = IR_MEM_BO(reg, addr_insn->op2);
+	ir_emit_load_mem(ctx, insn->type, def_reg, mem);
+
+	if (IR_REG_SPILLED(ctx->regs[def][0])) {
+		ir_emit_store(ctx, insn->type, def, reg);
```
henderkes:
```suggestion
		ir_emit_store(ctx, insn->type, def, def_reg);
```

ir_x86.dasc
```diff
@@ -13996,59 +14053,155 @@ static void ir_emit_lea(ir_ctx *ctx, ir_ref def, ir_type type)
...
+		mem = IR_MEM_O(addr_insn->op2);
+	} else {
+		ir_emit_tls_base_addr(ctx, reg, addr_insn->op1);
+		mem = IR_MEM_BO(reg, addr_insn->op2);
+	}
+
+	ir_emit_load_mem(ctx, insn->type, def_reg, mem);
+
+	if (IR_REG_SPILLED(ctx->regs[def][0])) {
+		ir_emit_store(ctx, insn->type, def, reg);
```
henderkes:
```suggestion
		ir_emit_store(ctx, insn->type, def, def_reg);
```

ir_x86.dasc
```diff
@@ -3451,6 +3477,25 @@ static bool all_usages_are_fusable(ir_ctx *ctx, ir_ref ref)
...
+	ir_ref n, *p;
+
+	if (use_list->count == 1 || (ctx->rules[ref] & IR_FUSED)) {
+		return 1;
+	}
+	n = use_list->count;
+	for (p = ctx->use_edges + use_list->refs; n > 0; p++, n--) {
+		ir_ref use = *p;
+		ir_op op = ctx->ir_base[use].op;
+		if (op != IR_LOAD && op != IR_LOAD_v && op != IR_STORE && op != IR_STORE_v) {
```
henderkes:
```suggestion
		if (op != IR_LOAD && op != IR_LOAD_v && ((op != IR_STORE && op != IR_STORE_v) || ctx->ir_base[use].op3 == ref)) {
```

don't fuse away a tls address that itself is being stored

ir_aarch64.dasc
```diff
@@ -1050,6 +1070,25 @@ static bool all_usages_are_fusable(ir_ctx *ctx, ir_ref ref)
...
+	ir_ref n, *p;
+
+	if (use_list->count == 1 || (ctx->rules[ref] & IR_FUSED)) {
+		return 1;
+	}
+	n = use_list->count;
+	for (p = ctx->use_edges + use_list->refs; n > 0; p++, n--) {
+		ir_ref use = *p;
+		ir_op op = ctx->ir_base[use].op;
+		if (op != IR_LOAD && op != IR_LOAD_v && op != IR_STORE && op != IR_STORE_v) {
```
henderkes:
```suggestion
		if (op != IR_LOAD && op != IR_LOAD_v && ((op != IR_STORE && op != IR_STORE_v) || ctx->ir_base[use].op3 == ref)) {
```

see above

ir.c
```diff
@@ -1069,14 +1070,16 @@ static ir_ref _ir_fold_cse(ir_ctx *ctx, uint32_t opt, ir_ref op1, ir_ref op2, ir
...
 
-		if (op1 > limit) {
-			limit = op1;
-		}
-		if (op2 > limit) {
-			limit = op2;
-		}
-		if (op3 > limit) {
-			limit = op3;
+		if ((opt & IR_OPT_OP_MASK) != IR_TLS_ADDR) {
```
henderkes: also needed in `ir_folding()`, before building `key`:

```c
if ((opt & IR_OPT_OP_MASK) == IR_TLS_ADDR)
	goto ir_fold_cse;
```

module and offset aren't ir refs

henderkes:
Created a PR targeting this to add some failing tests. Four should already be fixed by the inline comments. Need to sleep now.

dstogov:
Thanks for the feedback. It's very useful. I'll merge your fixes and will try to fix the remaining problems today.
I don't plan to support fusion of TLS_LOAD into other instructions yet (as you request in ``tls_addr_arithmetic.irt`` and ``tls_addr_update.irt``).

dstogov:
@henderkes I think, this is ready for another round of review.

I had to make ``TLS_ADDR`` not-foldable. Now it's linked into control list similar to old ``TLS``.
Otherwise GCM would hoist it above ``ENTRY`es.
After this change ``tls_addr_module.irt``, ``tls_addr_offset.irt`` and ``tls_addr_entry.out`` don't have a lot of sense.

henderkes:
The slop machine is satisfied, but I'll do a manual review later when I have my laptop.

ir_x86.dasc
```diff
@@ -2375,6 +2377,30 @@ get_arg_hints:
 				n = 1;
 			}
 			break;
+		case IR_TLS_LOAD:
+			insn = &ctx->ir_base[ref];
+			flags = IR_USE_MUST_BE_IN_REG;
+			if (insn->op1 >= 0 && !IR_IS_TYPE_INT(insn->type)) {
```
henderkes, line 2383:
```suggestion
			if (ctx->ir_base[insn->op2].op2 >= 0 && !IR_IS_TYPE_INT(insn->type)) {
```

We don't need to check the control instruction, we need to check the tls module operand to figure out if it's dynamic tls

ir_x86.dasc
```diff
@@ -2375,6 +2377,30 @@ get_arg_hints:
...
+			flags = IR_OP3_MUST_BE_IN_REG;
+			if (IR_IS_CONST_REF(insn->op3)) {
+				if (IR_IS_TYPE_INT(ctx->ir_base[insn->op3].type)) {
+					n = ir_add_const_tmp_reg(ctx, insn->op3, 3, n, constraints);
+				} else {
+					constraints->tmp_regs[n] = IR_TMP_REG(3, ctx->ir_base[insn->op3].type, IR_LOAD_SUB_REF, IR_USE_SUB_REF);
+					n++;
+				}
+			}
+			if (insn->op1 >= 0) {
```
henderkes, line 2399:
```suggestion
			if (ctx->ir_base[insn->op2].op2 >= 0) {
```

ir_x86.dasc
dstogov: Right. Thanks!

ir_aarch64.dasc
```diff
@@ -857,7 +857,8 @@ get_arg_hints:
 		case IR_TLS_LOAD:
 			insn = &ctx->ir_base[ref];
 			flags = IR_USE_MUST_BE_IN_REG;
-			if (insn->op1 >= 0 && !IR_IS_TYPE_INT(insn->type)) {
+			if (ctx->ir_base[insn->op2].op2 >= 0 /* we need an extra register to access dynamic TLS */
+			 && !IR_IS_TYPE_INT(insn->type)) {
```
henderkes: I don't think this works on aarch64, it needs a general register for the tls base even in static tls.

henderkes: (this is also the cause for the clang tailcall vm executing a lot more instructions compared the the gcc hybrid vm on aarch64, it's just that tls reloads happen more frequently which has a bigger impact on aarch64 than on x86)

dstogov: Yes. I saw the test failure. This is already fixed.

dstogov: PHP/JIT may load TLS address for the required module (or TCB) once per JIT trace (or function) and then perform all LOAD/STORES based on it.

henderkes:
https://github.com/dstogov/ir/pull/211 is the last thing, then I'm happy.

dstogov:
> #211 is the last thing, then I'm happy.

Please check the last commit

ir_x86.dasc
```diff
@@ -3508,15 +3508,22 @@ bool ir_may_fuse_tls_addr(ir_ctx *ctx, ir_ref ref)
...
+	if (use_list->count == 2 || (ctx->rules[ref] & IR_FUSED)) {
 		return 1;
 	}
 	n = use_list->count;
 	for (p = ctx->use_edges + use_list->refs; n > 0; p++, n--) {
 		ir_ref use = *p;
 		ir_op op = ctx->ir_base[use].op;
-		if (op != IR_LOAD && op != IR_LOAD_v
-		 && ((op != IR_STORE && op != IR_STORE_v) || ctx->ir_base[use].op3 == ref)) {
+		if (op == IR_LOAD && op == IR_LOAD_v) {
```
henderkes: did you mean `||`?

dstogov: oh. :(

henderkes:
Seems all good now!

PR #203, opened 2026-09-15
Run runtime tests at all optimization levels

jorgsowa:
Run every `tests/run` case at -O0, -O1, and -O2. Remove the runtime tests’ redundant per-case `-O0` arguments and include optimization levels in test failure summaries.

It exposed some bugs in all tests basically. What do you think about such approach?


dstogov:
The idea is interesting, but I would like to keep ``tester.c`` project independent.
It would be great to allow specification of few different --ARGS-- section (I'm not sure how this should look like), also may be several EXPECT sections (for different targets ans/or args), also it would be great to generalise TARGET into SKIPIF.

I won't merge this yet. PR for multiple ARGS would be great.

Anyway, thanks for finding ``-O0`` problems.
``-O0`` is really not a primary goal of the project (it's still incomplete), but I'll try to take a look..

dstogov:
Could you pleas re-run the CI on top of ``master``.
Two GUARD relayed tests ``.\tests\run\fcmp_005.irt`` and ``.\tests\run\fcmp_006.irt`` depend on optimized stack layout. Their failures with -O0 expected.

The other failures should be fixed in master.
Thanks!

dstogov:
Thanks.
``.\tests\run\argval_001.irt`` still fails on Windows-64. I'll take a look.
Everything else is expected.

PR #204, opened 2026-09-16
Fix GUARD reusing condition flags set in another basic block

iliaal:
`ir_match_insn()` folds `v = BINOP(a, b); c = CMP(v, 0); GUARD(c)` into `IR_GUARD_JCC_INT`, emitting the BINOP, skipping the CMP and branching on the flags the BINOP left behind. It only checked that the BINOP precedes the CMP in the IR. When GCM hoists a loop-invariant BINOP out of a loop, the jcc in the loop body reads flags the loop's own instructions have clobbered:

```
	movq %rsi, %rax
	addq %rdi, %rax        ; sets the flags the guard wants
	xorq %rcx, %rcx
.L1:
	js abort               ; reads the flags left by the cmpq below
	leaq 1(%rcx), %rcx
	cmpq %rdx, %rcx
	jl .L1
	retq
```

The fold now requires the BINOP in the guard's own block with only snapshots in between, the way `ir_match_fuse_load()` pairs `ir_in_same_block()` with `ir_match_has_mem_deps()`. `tests/x86_64/guard_001.irt` is the listing above; it fails before the change and passes after. `make test` reports 565 passed / 0 failed with the test, 564 / 0 without it.

Found from php/php-src#23693, where a recursive PHP function under the tracing JIT lost an `if ($newY < 0) continue;` and threw on an argument it had already proved non-negative.


dstogov (approved):
Excellent! :)

PR #206, opened 2026-09-16
Implement floating-point boolean NOT

jorgsowa:
Add floating-point boolean NOT support to the x86 and AArch64 backends.

Not sure it is correct direction, but it was marked in codebase as TODO in codebase, so I assumed it should be supported.


dstogov:
Oh sorry, that TODO was added long time ago. I would remove it now.

I think, we  don't need ``NOT`` for floating-point as it doesn't make any advantage against ``EQ(x, 0.0)``, and we already have fusion rules for ``IF(EQ(x, 0.0))`` and similar.

Anyway, now you know almost all details of code-generator.
I'll add review comments just to point to possible general improvements.
They just for knowledge...

ir_x86.dasc
```diff
@@ -2153,6 +2154,16 @@ get_arg_hints:
...
 			break;
+		case IR_BOOL_NOT_FP:
+			flags = IR_USE_MUST_BE_IN_REG | IR_OP1_MUST_BE_IN_REG;
+			insn = &ctx->ir_base[ref];
+			if (IR_IS_CONST_REF(insn->op1)) {
+				constraints->tmp_regs[0] = IR_TMP_REG(1, ctx->ir_base[insn->op1].type, IR_LOAD_SUB_REF, IR_USE_SUB_REF);
+				n = 1;
+			}
+			constraints->tmp_regs[n] = IR_SCRATCH_REG(IR_REG_RDX, IR_USE_SUB_REF, IR_SAVE_SUB_REF);
+			n++;
```
dstogov, line 2165: It's better to use not hard-coded temporary register.
```
constraints->tmp_regs[n] = IR_TMP_REG(2, IR_U32, IR_LOAD_SUB_REF, IR_DEF_SUB_REF);
```
Here ``2`` is an operand slot number where the register-allocator is going to store the assigned register. (0 - is for result, 1 - for op1, we can use 2 and 3 because NOT doesn't use them).
In the code generator routine you should use
```
ir_reg tmp_reg = ctx->regs[def][2]; 
```

dstogov (commented):
Everything else looks great.

jorgsowa:
Thank you for review. I will close this PR for now and take something else.

ir_x86.dasc
```diff
@@ -1611,6 +1611,7 @@ const ir_call_conv_dsc ir_call_conv_x86_fastcall = {
 	_(SMOD_PWR2)           \
 	_(BOOL_NOT)            \
 	_(BOOL_NOT_INT)        \
+	_(BOOL_NOT_FP)         \
```
jorgsowa, line 1614: @dstogov Should `BOOL_NOT_INT` be removed as well?

PR #209, opened 2026-09-17
Fold floating-point NOT to zero comparison

jorgsowa:
Implements alternative approach for floating-point boolean NOT explained in https://github.com/dstogov/ir/pull/206#issuecomment-5704226755


dstogov:
This doesn't make a lot of sense if we don't support code-generation for ``NOT(fp)``.
``NOT`` is actually a bitwise not (it shouldn't support fp operands).

jorgsowa:
Sure, I will close the PR

PR #210, opened 2026-09-17
Fix AArch64 FP div same-constant operands

jorgsowa:


tests/run/div_007.irt
```diff
@@ -0,0 +1,27 @@
+--TEST--
+DIV 007: div float constant by itself
+--ARGS--
+--run
```
dstogov: Without ``-O0`` optimizer will perform constant folding.
As result this test doesn't test the desired problem.

dstogov (changes requested):
The fix itself looks fine.
The test should be fixed.

Please check the same problem for other instructions (e.g.  ``MIN``, ``MAX``, ...)
You may do this in the next PR.

tests/run/div_007.irt
jorgsowa: That's why I think the change in test suite that test all optimizations level would be handy in such cases https://github.com/dstogov/ir/pull/203

I added back -O0 flag.

dstogov (approved):

